PR #4, opened 2019-11-07
CY-1486 - Export React components from Stage to new repository (part #3) 

qooban:


qooban:
@kubama Good catch, thanks. Anything apart from import added within https://github.com/cloudify-cosmo/cloudify-ui-components/pull/4/commits/08bf2370c58993b70067e2703f5ca92ca7c79326 ?

kubama:
@qooban Actually it seems it was the only one

PR #5, opened 2019-11-13
CY-1487 - Add library deployment mechanism

qooban:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "cloudify-ui-components",
-  "version": "1.0.0-alpha.1",
+  "version": "0.0.1",
```
tyacbovi: Did u mean to change the version?

qooban: Yes, I wanted to start with something `<1.0.0`, will be releasing it during development period.
When I finish it (= release `1.0.0`) I will remove all the versions below 1.0.0 from NPM registry.

PR #16, opened 2019-12-16
Added loki visual regression testing

kubama:


qooban:
@kubama I see you are trying to run Storybook server in CI, but maybe it is not necessary and you can use that solution: https://loki.js.org/continuous-integration.html ?

kubama:
@qooban Yes, it was the way to go

package.json
```diff
@@ -101,5 +103,17 @@
     "react-dom": "^16.10.2",
     "semantic-ui-react": "^0.84.0",
     "semantic-ui-css": "^2.4.1"
+  },
+  "loki": {
```
qooban: General comment about screenshots:
1. `Loading`, `LoadingOverlay` - they are animated. How the diffing engine is handling that?
2. `Date` inputs, `Modal`s, `Popup`s - it would be good to open them before taking screenshot (or even better take two screenshots - close/open).

package.json
```diff
@@ -23,6 +23,7 @@
     "buildDev": "babel ./src --ignore '**/*.stories.jsx' --copy-files --delete-dir-on-start --out-dir es --watch",
     "lint": "eslint --ext js,jsx src test",
     "test:only": "jest test --coverage",
+    "test:loki": "build-storybook && loki --chromeTolerance 101 --requireReference --reactUri file:./storybook-static",
```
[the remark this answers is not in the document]
kubama: It doesn't seem to be percentage as the documentation states.
101 is the lowest value for this parameter that allows to pass comparison on CI due to environment specific font rendering differences. At the same time I tested that it is still possible to introduce changes that would cause comparison to fail with this setting.

package.json
```diff
@@ -23,6 +23,7 @@
...
     "test": "npm run lint && npm run test:only",
```
qooban: Let's add `test:loki` here.

This script is used when publishing package to npm.

kubama: Done

package.json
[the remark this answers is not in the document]
qooban: OK, so what is passed to graphicksmagic? (I assume this diffing engine is used by default)

Check: 
- https://loki.js.org/configuration.html#gm-https-githubcom-aheckmann-gm
- https://github.com/aheckmann/gm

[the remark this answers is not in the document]
kubama: gm is passed two images - reference image and actual image

package.json
kubama: 1. It compares first animation frames
2. The only way to do this would be to have a story with the component initially open - for modals it's not really an option, but for Date and popups I'll try to extend components with props to make it possible.

PR #28, opened 2020-02-17
Switched from semantic-ui-css to semantic-ui-css-offline

kubama:


kubama:
@qooban In fact this is just a change in peerDependencies - how about postponing the release until something relevant is merged?

qooban:
@kubama If you won't get blocking errors during `npm install` in cloudify-stage and cloudify-blueprint-composer, I'm OK with postponing it.

PR #33, opened 2020-07-01
CY-3014 add new jenkins pipeline

warolv:
CY-3014 add new jenkins pipeline


warolv:
> Looks fine 👍
> I think you can remove CircleCI configuration for the same reason as in UI Common.

I will disable a webhook meantime, and in a couple of sprints will remove config also. Will remove after optimization.

PR #49, opened 2020-11-18
RD-537 - Add support for logoUrl and showVersionDetails parameters in theme

qooban:
* Added support for logoUrl and showVersionDetails theme parameters to simplify layout components usage in Stage and Composer
* Added `className` prop (with default values) to layout components to simplify its customization


src/components/layout/HeaderBanner/HeaderBanner.jsx
```diff
@@ -13,27 +14,53 @@ import LicenseEdition from './LicenseEdition';
...
+
+    let url = logoUrl;
+    if (theme && theme.logoUrl) {
+        url = theme.logoUrl;
+    }
+
+    let showDetails = showVersionDetails;
+    if (theme && _.isBoolean(theme.showVersionDetails)) {
+        showDetails = theme.showVersionDetails;
+    }
```
kubama, line 46: I'd use `{}` as default value for `theme` and then:
```suggestion
    const color = theme.headerTextColor || colors.white;
    const url = theme.logoUrl || logoUrl;
    const showDetails = _.isBoolean(theme.showVersionDetails) ?  theme.showVersionDetails : showVersionDetails;
```

src/components/layout/Logo/Logo.jsx
```diff
@@ -1,12 +1,29 @@
...
  */
 export default function Logo({ className, style, url }) {
+    const theme = useContext(ThemeContext);
+    let logoUrl = logo;
+    if (theme && theme.logoUrl) {
+        logoUrl = theme.logoUrl;
+    }
+    if (url) {
+        logoUrl = url;
+    }
```
kubama, line 26: This logic is not in line with what is described above

qooban: Fixed

src/components/layout/HeaderBanner/HeaderBanner.jsx
qooban: Makes sense. Fixed.

kubama (approved):

PR #54, opened 2021-01-19
RD-1067 Scroll only modal content

Gelio:
I fixed the modal scrolling by fiddling with `display: flex`, `overflow: auto` and setting correct element heights. The current state of things (some part of the video comes from cloudify-stage):

Full screen:

https://user-images.githubusercontent.com/889383/105055414-547ab980-5a73-11eb-8bab-d33c94b9eefc.mp4

On smaller screen sizes:

https://user-images.githubusercontent.com/889383/105173166-ed1b4300-5b20-11eb-917e-2263c35a827d.mp4

Main changes:

1. Add a new `Modal` component (a re-export of Semantic-UI's Modal with correct styles)
2. Use that component in other stories that require modals
3. Add stories that show how the modal behaves when it has long content
4. Update storybook to the latest version
5. ~Build storybook in a special `VISUAL_TESTING` mode.~ 

    ~In that mode, modals will be opened automatically. This way loki tests the modal content, and not the button that opened the modal (as it does now)~ (this issue cannot be solved right now, since modals mount themselves in in `document.body`, and if I wanted to change so they mount inside the story, I would use `mountNode` and a `ref`, yet that does not work in storybook ATM https://github.com/storybookjs/storybook/issues/10446

    I've added https://cloudifysource.atlassian.net/browse/RD-1146 to track it

This is only a part of the changes for the ticket. Changes are also required in cloudify-stage (and maybe cloudify-blueprint-composer, if we do not split the ticket into 2).


package.json
```diff
@@ -43,11 +43,11 @@
...
     "@babel/preset-react": "^7.12.10",
-    "@storybook/addon-docs": "^6.0.27",
-    "@storybook/addon-storyshots": "^6.0.27",
+    "@storybook/addon-docs": "^6.1.14",
+    "@storybook/addon-storyshots": "^6.1.14",
     "@storybook/preset-scss": "^1.0.3",
-    "@storybook/react": "^6.0.27",
-    "@storybook/theming": "^6.0.27",
+    "@storybook/react": "^6.1.14",
+    "@storybook/theming": "^6.1.14",
```
kubama, line 50: How are these upgrades related to the ticket's scope?

kubama (commented):
This is for sure much better than it used to be 👍

package.json
Gelio: They are not :slightly_smiling_face: I updated the packages just to keep up-to-date (and also due to a Storybook popup asking for an upgrade)

kubama: We use "squash and merge" when merging to master - if I was looking at the git history after the merge I would mistakenly think that these changes are internal part of work done as part of this ticket and that they were actually required for some reason. These are just `devDependencies` so in this case it may be not that important and of course I'm fine with these particular upgrades, but in general I'd avoid smuggling any unrelated changes in PRs.

kubama (approved):

package.json
Gelio: I see, I did not think about it while making the changes. I'll consciously consider it next time when coming to such a scenario

Gelio:
Thanks @kubama for the review

Looks like the visual regression tests failed, and rightly so, since the appearance of the components changed. I'll try to inspect how to update those

src/components/modal/Modal/Modal.css
```diff
@@ -0,0 +1,12 @@
...
+    display: flex !important;
+    flex-direction: column;
+    /* NOTE: 2rem come from vertical margin, 2rem come from parent's padding */
+    max-height: calc(100% - 4rem);
+}
+
+.ui.scrolling.modal .content {
+    flex-shrink: 1;
+    overflow: auto;
+}
```
qooban, line 12: I assume all is fine on smaller screens with those changes?

src/components/modal/Modal/Modal.stories.jsx
```diff
@@ -0,0 +1,44 @@
...
+    title: 'Modal/Modal',
+    component: Modal,
+    parameters: {
+        docs: {
+            description: {
+                component: `Semantic-UI's Modal component with fixed content scrolling.
+                    See [Semantic-UI documentation](https://semantic-ui.com/modules/modal.html#/definition) for more information`
+            }
+        }
+    }
```
qooban, line 17: Hmm... Isn't that part automatically taken from the comment in component's code?
I assume not probably because we directly export it and storybook does not handle it?

Anyways, let's keep it only in one place and remove it from the other not to have it duplicated.

src/components/modal/Modal/Modal.stories.jsx
```diff
@@ -0,0 +1,44 @@
...
+                <Modal.Header>Header</Modal.Header>
+                <Modal.Content>
+                    <p>Content, possibly very long.</p>
+
+                    <Dropdown options={[{ text: 'Example 1' }, { text: 'Example 2' }]} selection />
+
+                    {Array.from({ length: 30 }).map((_, index) => (
+                        // eslint-disable-next-line react/no-array-index-key
+                        <p key={index}>A paragraph of text, only to expand the modal so it scrolls</p>
+                    ))}
```
qooban, line 36: Yeah, thanks for adding that :-) So if we can just change the order (first `Array`, then `Dropdown`), we would have the exact test of the thing I was worried about.

Btw I found related PR: https://github.com/cloudify-cosmo/cloudify-stage/pull/498
It was fixed then, I think, and Semantic-UI-React version was quite old, so we should not have any issues with this.

src/components/modal/ReadmeModal/ReadmeModal.stories.jsx
```diff
@@ -1,51 +1,89 @@
...
+    <li>Aliquam tincidunt mauris eu risus.</li>
+</ul>
+
+<pre><code>
+#header h1 a {
+    display: block;
+    width: 300px;
+    height: 80px;
+}
+</code></pre>`
```
qooban, line 67: I'm just wondering how difficult it would be to move it to a separate file (real HTML file) and import it?
As storybook uses webpack, maybe that would just be like simple file import. 

Anyways, don't spend to much time on it if it is not easy change.

qooban (commented):
I like the change, looks better than it was 👍🏻 

Added some minor comments.

src/components/modal/Modal/Modal.css
Gelio: Yes :smile: I'm sorry about not including a recording for smaller screen sizes. I have included it in the PR's description :slightly_smiling_face:

src/components/modal/Modal/Modal.stories.jsx
Gelio: You are correct, storybook does not detect the JSDoc comment since it's not a new component, but a re-export, hence I added it here for the storybook description.

I thought I may as well leave it in the main component to clearly state why it's needed, but I can see it may be inferred from the context (i.e. the css file that's included) :smile: I'll remove it

src/components/modal/Modal/Modal.stories.jsx
Gelio: Good catch! Turns out there was a problem :smile: 

image

Let me fix it

Gelio: Weird... That dropdown expands the content of the modal when it's open:
image

src/components/modal/ReadmeModal/ReadmeModal.stories.jsx
Gelio: Good idea! It was a matter of using the `raw-loader` to import the HTML file (the content of the modal looked more like HTML than MD) :smile: I did not change storybook's webpack config, but decided to use the loader just in that one-place, which eslint did not like

src/components/modal/Modal/Modal.stories.jsx
Gelio: After a bit of research, that problem is quite well known and without an easy CSS fix. To behave properly it requires making the dropdown positioned absolutely with regards to the whole modal, not to the trigger, and then adding some JS to track the scroll position and update the absolutely positioned dropdown accordingly :worried: 

I'd rather not spend time on it now. The workaround here would be to make the dropdown a dropup :smile: so it expands upward (there's an `upward` prop)

Some resources I found:
1. https://css-tricks.com/popping-hidden-overflow/
2. https://stackoverflow.com/questions/47930194/drop-down-menu-hidden-by-div-with-overflow-auto
3. https://github.com/Semantic-Org/Semantic-UI/issues/4495

qooban: Hmm... Yeah, I know about `upward` prop, AFAIR we tried this as well. That solves issues with dropdowns explicitly added to the modal. But there still might be an issue as we can have dynamically added dropdowns to the modals AFAIR (eg. in `DeployBlueprintModal` - https://github.com/cloudify-cosmo/cloudify-stage/blob/71a132f726cd05ac0952ec019a61a745da71ed32/widgets/common/src/DeployBlueprintModal.jsx#L421).

Gelio:
The PR is ready for the final review. I have updated loki's reference files, so CI finally passes

I've updated the description with the latest changes

kubama (approved):

qooban (commented):
I've got just one concern before merging it. I described it already [here](https://github.com/cloudify-cosmo/cloudify-ui-components/pull/54#discussion_r560999107).

Could you test it deeply? Dropdown should look properly.

1. Let's check eg. `DeployBlueprintModal` - there are blueprint and site name dropdowns. What if there are lots of blueprints/sites to display?
2. In the same modal you can have dynamic content (depending on the blueprint, inputs provided by the user to the blueprint can be different, it can happen that dropdown will be at the bottom, should we make all dropdowns using upward, then?)
3. What if there is a very simple modal with just one dropdown (with long list) inside? I understand that your suggestion is to use `upward` then. How will it look like? Will the dropdown list be presented on top of the modal header then?

src/components/modal/Modal/Modal.stories.jsx
Gelio: The problem is quite tricky.

I did not find a way for the Dropdown to appear on top of the bottom actions (https://github.com/Semantic-Org/Semantic-UI/issues/1340, https://github.com/Semantic-Org/Semantic-UI-React/issues/497#issuecomment-463091039 (I even found your comment :smile: ))

I have some ideas for possible solutions, though. See below

## Solution A - a hack

A somewhat of a hack would be to:

```ts
<Dropdown
  onOpen={() => {
    if (dropdownIsInsideModal && contentHeightExpandedAfterOpeningDropdown) {
      scrollModalToTheBottom();
    }
  }}
/>
```

That could be unexpected for the user, though

## Solution B - limited scope

I found out about the `scrolling` prop on `Modal.Content` (https://react.semantic-ui.com/modules/modal/ - see the Scrolling Content example). Thus, we could only add that prop for the to-date modals that cause trouble with closing - the readme modal, and the `BlueprintSources` modal. That does not solve the problem for any other modal whose content is too tall for the viewport, e,g. that `DeployBlueprintModal`.

Small viewport, action buttons are not visible:
image

Dropdowns work well in this scenario, as they open upward then close to the bottom of the modal:
image

## Solution C - ignore the problem for dynamically-generated dropdowns

We could leave the current solution I prepared - all modals are scrollable **when their content is too tall to fit in the viewport**. Notice the bolded condition. Semantic-UI only adds the `scrolling` class to the modal when the content is too tall.

This means that modals that fit the content will not scroll, and the dropdowns will appear on top of actions:
image

Thus, in this particular modal (`DeployBlueprintModal`), there's no need for an `upward` prop for most cases. One problematic case here would be if the user's viewport is quite short. Then, the modal is scrollable, and the user would have to scroll to see the full dropdown content:
image

However, this has a huge benefit IMO - the user can always see the header and the actions, giving him additional context.

Responding to your questions from https://github.com/cloudify-cosmo/cloudify-ui-components/pull/54#pullrequestreview-574020512:
1. It looks fine. If you can fit the initial size of the modal on your viewport, then the modal is not scrollable, which makes the dropdown appear on top of actions: 
    image
2. If there's dynamic content, there's little we can do. We consciously ignore this situation. Does it happen often that it's actually the dropdown that appears last?
3. It looks fine without the `upward` prop:
    image

    image

    The key point here is that the modal becomes scrollable and *eats* the dropdown only if the initial content of the modal (before opening the dropdown) is too tall.

Just as an experiment, this is how a final `Dropdown` with an `upward` prop looks at the bottom of the content:
image


## My recommendation

I would say solution C is the most UX-friendly. The only case it does not handle is if you have a long modal, and at the bottom of it is a dropdown, and that dropdown is generated dynamically (for hand-crafted dropdowns we can use `upward`). All other cases seem to be covered for me.

A benefit of this approach is always displaying the header and the actions, which lets the user easily orient themselves (similarly to how Stage always displays a header and a sidebar, and does not scroll).

If not C, then B is probably safest, although I would still encourage going with C.

## Summary

Let me know what you think :smile:

qooban: Thank you very much for that great analysis and summary!

Reg. case no. 2 from my list I was more afraid about dropdowns coming from the deployment inputs (which depends on the blueprint), but... Yeah, that should not very often be the case, I think.

So I'm OK with the solution you suggested.

qooban (approved):

PR #55, opened 2021-01-20
RD-1111 - Update react-json-view to fix security issues

qooban:


Gelio:
For some reason, `jest` collected the coverage that is somehow defined in the `react-json-view` library in `node_modules` :exploding_head: I already filed an issue there, see https://github.com/mac-s-g/react-json-view/issues/317

To get this PR to pass, I tried to prefix each entry in `collectCoverageFrom` with `<rootDir>`, so it looks like 
```ts
    collectCoverageFrom: [
        '<rootDir>/src/**/*.jsx',
        '<rootDir>/src/**/*.js',
        '!<rootDir>/src/**/*.stories.jsx',
        '!<rootDir>/src/decorators/**'
    ],
```

but to no avail. I tried to exclude `node_modules` (`'!node_modules'`) but that also did not work :slightly_frowning_face: I also tried upgrading jest to the latest version (26.6.3). The tests passed, but the coverage check failed :smile: I tried specifying `coveragePathIgnorePatterns` too - no success 

I'm slightly out of ideas for now

qooban:
@Gelio Thanks a lot for the investigation! Looks like during the meeting the issue you reported on react-json-view side (https://github.com/mac-s-g/react-json-view/issues/317) was fixed. I'll update, check PR and soon prepare for the review.

Gelio:
I did an `npm install` and I got some minor diffs in `package-lock.json` :smile: 

image

I'm certain this does not matter, but will probably cause unnecessary diffs when I upgrade something later. I'm using npm v6.14.11. Which one are you using? I reckon it would be best if we used the package-lock generated by the most recent version

Gelio (approved):
Looks good :+1: I've tested that it works in storybook

qooban:
@Gelio I'm using the same version of `npm`. 
BTW I'm using `npm ci` to get exactly the same packages installed.

Gelio (approved):
Then I'm not sure why there was a difference :smile: Anyway, looks good

PR #56, opened 2021-02-01
RD-1230 Add TypeScript

Gelio:
This PR adds TypeScript support to the repository. It depends on https://github.com/cloudify-cosmo/cloudify-ui-common/pull/46 to be merged and released first

Bundle size comparison:

Repository | File | Size before TS | Size after TS
-- | -- | -- | --
cloudify-ui-components | es/**/* | 1146 | 1376

The difference probably results from the fact that JSX was transpiled to `React.createElement` now

## Changes
1. Storybook
    1. Adjust paths so it picks up TSX files too
2. Babel
    1. Add necessary preset to support TS syntax
3. Jest
    1. Uses `babel-jest`, so support for TS is automatic
    2. Include TS extensions in coverage collecting
4. Unify the output (`dist` dir) of `npm run build` and `npm run buildDev`
5. Add npm scripts for generating type declarations (`typings`) and checking the types
6. ESLint
    1. Include TS extensions
7. `tsc` will report compilation errors for TS files only. This means that JS files can contain invalid types, but once we gradually move them to TS, they should be 100% valid
8. Fix existing linting errors in some components (related to the order of methods in the class)
9. A single component (`Form.ColorPicker`) uses TypeScript now. Its tests and stories are also in TypeScript.

    While modifying the stories for that component, I used storybook args, which yields a pretty nice UI:
    
    image


src/components/form/DateRangeInput/DateRangeInput.jsx
```diff
@@ -96,35 +96,6 @@ export default class DateRangeInput extends React.PureComponent {
         }
     }
 
-    getTimeFilterObject() {
```
Gelio, line 99: There was an existing ESLint error about the order of methods in the class being invalid. I only moved the methods below, not modifying anything else

kubama (dismissed):

Gelio:
@qooban @kubama @gregcfy I've updated `cloudify-ui-common` in this PR, meaning that it is ready to be merged :slightly_smiling_face:

Gelio:
After merging this PR I plan on publishing a **minor** version of `cloudify-ui-components`. I did not think of any breaking changes for the consumers of this package. The output should stay as it was, and the added feature is having type declarations and declaration maps, hence I'd say it would be a minor version bump :slightly_smiling_face: 

Keep in mind that this would mean that the major version of `cloudify-ui-common` and this repo will be unsynchronized. I don't think that's a problem, but I wanted to bring it up :slightly_smiling_face:

src/components/form/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * Hexadecimal color value
+     */
+    value?: string;
+
+    /**
+     * Function called on color change
+     */
+    onChange?: (event: ChangeEvent<HTMLInputElement>, data: { name: string; value: string }) => void;
```
qooban, line 20: Definitely it's much more clear what are the arguments right now 👍🏻

src/components/form/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,39 @@
...
+}
+
+/**
+ * ColorPicker is a component showing CompactPicker component from [react-color](https://casesandberg.github.io/react-color/) library
+ *
+ * Accessible as `ColorPicker` or `Form.ColorPicker`.
+ *
+ */
+const ColorPicker: FunctionComponent<ColorPickerProps> = ({ name = '', value = '#000000', onChange = () => {} }) => {
```
qooban, line 29: Nice, so `name = '', value = '#000000', onChange = () => {}` serves the same way as `defaultProps`?

src/components/form/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,39 @@
+import React, { ChangeEvent, FunctionComponent } from 'react';
+
```
qooban, line 2: Very, very minor thing - probably we don't need extra empty line here, right?
I think it make sense when we divide internal from external imports.

src/components/form/ColorPicker/ColorPicker.jsx
```diff
@@ -1,44 +0,0 @@
-import React from 'react';
```
qooban, line 1: I'm wondering how to handle those changes (.jsx -> .tsx) without losing history. I remember we had discussions with @kubama about it and in some cases even if you mark the file as moved it is difficult to "convince" git that it is really file move and not file deletion. Do you know any methods for that?

qooban (commented):
Changes seems quite clear for me 👍🏻 I think we are going in the right direction. 
Left some minor comments/questions.

Re:
> Keep in mind that this would mean that the major version of `cloudify-ui-common` and this repo will be unsynchronized. I don't think that's a problem, but I wanted to bring it up 🙂 

By unsynchronized you mean that we are doing major bump in `cloudify-ui-common` (breaking backward compatibility) and here in `cloudify-ui-components` we are planning just minor version bump with changed dependency to `cloudify-ui-common` (new 3.0.0)? I think that is fine and acceptable.

src/components/form/ColorPicker/ColorPicker.tsx
Gelio: I'm glad :smile: and I definitely agree! TS for the win :smile:

src/components/form/ColorPicker/ColorPicker.tsx
Gelio: Yes, exactly :slightly_smiling_face: It's just the default and ES-native way of specifying the default value of some parameter, and it works with destructuring too. It also works for regular functions, not only components :slightly_smiling_face: 

In essence, it is either one of these, depending on the use case (it's the 2nd one in the case of destructuring props):
1. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters
2. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Default_values_2

src/components/form/ColorPicker/ColorPicker.tsx
Gelio: That's right :smile: There was an empty line in the original file and I did not think much about it, but right, I'll fix it :slightly_smiling_face:

qooban:
@Gelio One more question about "Bundle size comparison": I did a quick check and it seems to me that we don't have such a big difference. Am I missing something?

`npm ci && npm run build && du -bs es` executed on two different branches:
- for `RD-1230-add-typescript`: 460274
- for `master`: 389499

When I looked at the differences, it looks like with the new build we include Storybook stories, which shouldn't be included IMHO. Looks like stories removal line in `build-components.sh` needs to be updated. If you remove stories, the size should be quite similar.

src/components/form/ColorPicker/ColorPicker.tsx
qooban: Yep, that (ES native way, regular functions) I was aware of. I was only not sure how it works with default props. Thanks 👍🏻

src/components/form/ColorPicker/ColorPicker.jsx
Gelio: Well, yeah, I reckon I can try having 2 separate commits for each move.
1. Rename the file
2. Modify the file

According to https://stackoverflow.com/a/433114, this should allow git to see that as a rename. I will try to do that, but that will require rewriting history via an interactive rebase :slightly_smiling_face:

qooban: Thanks for checking that 👍🏻 I leave to your decision whether to do it or not within this PR. I was more concern about the future changes. This (ColorPicker) component is not widely used and it is simple, so we actually don't lose much.

Gelio:
@qooban 
> By unsynchronized you mean that we are doing major bump in cloudify-ui-common (breaking backward compatibility) and here in cloudify-ui-components we are planning just minor version bump with changed dependency to cloudify-ui-common (new 3.0.0)? I think that is fine and acceptable.

Yes, by desynchronizing I mean having 3.0 for common and 2.3 for components (different major versions).

>When I looked at the differences, it looks like with the new build we include Storybook stories, which shouldn't be included IMHO. Looks like stories removal line in build-components.sh needs to be updated. If you remove stories, the size should be quite similar.

Good idea :slightly_smiling_face: I've implemented it, and now the differences are minimal:
* 155962 for this branch
* 156322 for `master`

src/components/form/ColorPicker/ColorPicker.jsx
Gelio: Ok then :slightly_smiling_face: I've made the changes anyway (I have split 730bde3ad1c108ac19a4b50372c70fe65d3455d7 into 2 commits, one that just renames the files, and the 2nd one that makes changes).

It did not seem to work, though.

You can see that https://github.com/cloudify-cosmo/cloudify-ui-components/pull/56/commits/8adfc38321ce9d495e6fcd4a1649da1c2fad8aa2 renames the files, and https://github.com/cloudify-cosmo/cloudify-ui-components/pull/56/commits/6a40b637693eb8c08801dd3324276e0806dc74ae makes changes to those files, but in the main diff, the diff still shows removing one file and adding a new one :worried: 

We can try to keep doing that pattern (2 commits instead of 1) to hope that it will have some benefits in the end :slightly_smiling_face:

Gelio:
The recent force-push is related to https://github.com/cloudify-cosmo/cloudify-ui-components/pull/56#discussion_r572850587

src/components/form/ColorPicker/ColorPicker.jsx
qooban: Thanks for checking that 👍🏻 
It seems that we had the same problem.

Gelio: Haha yeah, I guess that's karma :smile:

Gelio:
@qooban @kubama Are you planning on re-reviewing this PR? I have addressed all comments and I'm not clear what's the status of your reviews :slightly_smiling_face:

kubama (approved):

qooban:
@Gelio Sure, will do, probably tomorrow. For the future: let us know by clicking re-review icon, that triggers mail notification (and at least for me it is a sign that PR is ready for re-review).

Gelio:
@qooban Right, I assumed that you resolving all the discussions when I responded to each one is sufficient, but you're right, I did not click the re-request button :slightly_smiling_face: Thanks for a reminder

@kubama Thanks for the review :slightly_smiling_face:

qooban (approved):
Looks good 👍🏻

PR #57, opened 2021-02-15
RD-1398 - Fix Checkbox and FieldLabel prop types

qooban:


kubama (commented):
I think it's a little bit unexpected to support node but not string type for a description/help property. I think we could easily support both string and node.

qooban:
@kubama According to https://reactjs.org/docs/typechecking-with-proptypes.html - `string` is part of the `node` type, so both are supported:
```
  // Anything that can be rendered: numbers, strings, elements or an array
  // (or fragment) containing these types.
  optionalNode: PropTypes.node,
```
We are already passing React elements to `help` prop, so that change is only to make React aware that it is valid and not throw error in the browser console for something which works fine. 

See that `PopupHelp` (used by `FieldLabel`) is already using `PropTypes.node` as an expected type of `content` prop.

kubama:
> @kubama According to https://reactjs.org/docs/typechecking-with-proptypes.html - `string` is part of the `node` type, so both are supported:

Cool, I didn't now that.

kubama (approved):

Gelio:
Too bad I'm late to this PR to suggest using TS for those components :smile: They are quite small, so it would have been a good place :smile: 

Anyway, good job fixing it with the correct type :slightly_smiling_face: FYI, in TS the type corresponding to `PropTypes.node` is `ReactNode` :slightly_smiling_face: that's if you were to transform some other component in the future and tried to look for such a type

PR #58, opened 2021-02-24
RD-1222 Adjust DataTable to support basic functionalities for Environments

Gelio:
This PR adjusts the `DataTable` component and related components to add functionalities necessary for the Deployments View widget.

The necessary functionalities are:
1. `tooltip` for `TableColumn`
2. `colSpan` for `TableDataCell`

Moreover, in this PR I have migrated some files to TypeScript. Since git does not show renamed in the full diff, I suggest reviewing the PR commit by commit, as the diffs for individual commits show renamed correctly :slightly_smiling_face: 

The PR also fixes an error in type-checking tests when using a non-relative path. Moreover, I have removed `checkJs` overrides as they are no longer necessary with the `cloudify-ui-common` upgrade.

This PR relies on changes to ESLint config done in https://github.com/cloudify-cosmo/cloudify-ui-common/pull/50

Stories for the new functionalities:

https://user-images.githubusercontent.com/889383/109131741-5a8b4680-7753-11eb-937f-ab5ec7878d80.mp4

## To do

- [x] Add stories for the added functionalities
- [x] Remove redundant `eslint-disable` comments for rules disabled in https://github.com/cloudify-cosmo/cloudify-ui-common/pull/50
- [x] Update loki screenshots for `DataTable` stories


Gelio:
I see the errors in tests, I will take care of them shortly :slightly_smiling_face:

src/components/data/DataTable/TableColumn.tsx
```diff
@@ -1,17 +1,64 @@
-import React from 'react';
+// Default props are defined when destructuring
+/* eslint-disable react/require-default-props */
```
kubama, line 2: I don't think it is a good practice - one could add a new optional prop without default value in props destructuring and it would be unnoticed as there is no eslint rule to check that.

Gelio: Actually, I would say avoiding `defaultProps` is the recommended practice, as they are going to be deprecated :slightly_smiling_face: https://github.com/reactjs/rfcs/pull/107

You do not need `defaultProps` when you're using TypeScript, because if you try to use an optional prop without specifying a default value when descructuring, you will get a compilation error :slightly_smiling_face: Try it yourself

kubama: Cool, I wasn't aware of this.
But if so, why not to disable this rule globally for all tsx files?

Gelio: Yeah, good point, considering it for a while, to my mind it does make sense :slightly_smiling_face: 

I have another rule that I would like to disable (https://github.com/typescript-eslint/typescript-eslint/blob/v4.14.1/packages/eslint-plugin/docs/rules/no-use-before-define.md is superseded by TypeScript), thus to avoid adding unnecessary `eslint-disable`s, I will raise a PR cloudify-ui-common to adjust the config :slightly_smiling_face:

Gelio: Done, see https://github.com/cloudify-cosmo/cloudify-ui-common/pull/50 :slightly_smiling_face:

Gelio:
Since I have changed the stories, loki will probably fail, so I still need to update the screenshots after it fails

src/components/data/DataTable/TableColumn.tsx
```diff
@@ -0,0 +1,111 @@
...
+export default TableColumn;
+
+TableColumn.propTypes = {
+    label: PropTypes.node,
+    name: PropTypes.string,
+    width: PropTypes.string,
+    show: PropTypes.bool,
+    tooltip: PropTypes.node,
+    centerAligned: PropTypes.bool
+};
```
qooban, line 101: Another dumb question. Looks like strange overhead for me - you need to define types in `TableColumnProps` and then again define it similarly here using `PropTypes`. Is it temporary? Or is it how we will need to do that in TypeScript?

qooban (commented):
Added two questions.

src/components/data/DataTable/TableDataCell.tsx
```diff
@@ -0,0 +1,44 @@
...
+export default TableDataCell;
+
+TableDataCell.propTypes = {
+    // @ts-expect-error Children props is provided by default
+    children: PropTypes.node,
+    className: PropTypes.string,
+    rowSpan: PropTypes.number,
+    colSpan: PropTypes.number,
+    // eslint-disable-next-line react/forbid-prop-types
+    style: PropTypes.object
```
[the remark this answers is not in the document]
Gelio: I didn't notice a native `PropType` for `CSSProperties`. We could almost go with 

```suggestion
    style: PropTypes.shape<CSSProperties>({})
```

but there's some type error, so it cannot be done in the end

https://github.com/brigand/react-style-proptype is a package that exposes a proptype for style, but I don't think it's worth installing at the moment :smile:

src/components/data/DataTable/TableColumn.tsx
Gelio: I wanted to maintain backward compatibility with the validation that PropTypes provide in runtime, hence I left `propTypes` be :slightly_smiling_face: This is so that the validation can still happen for consumers that do not use TypeScript and will not see the compilation errors due to props mismatches

Ideally, we would be strictly only using TypeScript, and then, PropTypes are obsolete, as you get the same benefits of declaring the types, but during compilation time, resulting in no runtime overhead and no duplication (the duplication we have now - both types and propTypes)

I suggest we remove `propTypes` once we use TypeScript for all components, which is probably pretty far in the future :smile:

Gelio:
Since I added stories, I had to update test snapshots. The full diff is not great, so I encourage reviewing https://github.com/cloudify-cosmo/cloudify-ui-components/pull/58/commits/067b1964406675ac76edd706f7b7bc0c1fee7e5a and https://github.com/cloudify-cosmo/cloudify-ui-components/pull/58/commits/82262896f1bebcedf1c23d861224b64402e9322f individually :slightly_smiling_face:

kubama (approved):

src/components/data/DataTable/TableDataCell.tsx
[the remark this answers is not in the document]
qooban: OK, got it 👍🏻 All right then, let's leave it as it is. 

BTW there is small trick for such cases to avoid disabling the rule `react/forbid-prop-types`.
Just use:
```suggestion
    style: PropTypes.shape({})
```
🙂

src/components/data/DataTable/TableColumn.tsx
qooban: Got it, clear now 👍🏻

qooban (approved):
Looks good 👍🏻

Gelio:
Thanks for reviews @qooban  and @kubama ! I'm merging the PR :slightly_smiling_face: I'll wait for #59 before releasing a minor version of the components package :slightly_smiling_face:

src/components/data/DataTable/TableDataCell.tsx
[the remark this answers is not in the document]
Gelio: Oh, I see :smile: Haha if we are doing such _tricks_, maybe we should disable the rule entirely :smile: ? Or reconfigure it so we don't need such workarounds

Anyway, that's out of scope, just saying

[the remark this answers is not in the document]
qooban: 😄 I know it's not perfect 🙂 , but IMO:
1. The rule make sense, it encourage us to write more details about the expected type.
2. That _trick_ is useful and is accepted in such obvious cases when the type is very broad and commonly known (like `style` prop). In other cases I always try to fill the details of the expected object.

I think with those assumptions above it make sense to keep the rule.
But as always - there is human factor and depends on our common sense to keep the right balance (very strict rules vs. developer responsibility) 🙂

PR #65, opened 2021-03-15
CY-2513 Use Roboto Regular font in Storybook's sidebar

Gelio:
This PR fixes [CY-2513](https://cloudifysource.atlassian.net/browse/CY-2513).

I could not find a way to allow importing the `theme` file and use the font defined there, similarly to how it is done in `theme.js` (https://github.com/cloudify-cosmo/cloudify-ui-components/blob/e654e010d19778c1ab5831c0d29195b6e4ac8703/.storybook/theme.js#L3-L5)

since SCSS does not seem to be handled correctly in `manager.js` (yet it is handled correctly in `preview.js`). I presume the loading of those 2 files (`manager.js` and `preview.js`) happens in separate places. Thus, `preview.js` can use SCSS (registered in https://github.com/cloudify-cosmo/cloudify-ui-components/blob/e654e010d19778c1ab5831c0d29195b6e4ac8703/.storybook/main.js#L8), whereas `manager.js` cannot.

I have tried importing the SCSS file directly, specifying the loaders in the import path (`!style-loader!css-loader!sass-loader!cloudify-ui-common/...`), but none worked. The error:
image

## Screenshots

Before:
image

After:
image

---

Overall, I'm not certain about using this font here :smile: It looks odd to me. Maybe I have gotten accustomed to the previous one, though.

### DIFF
--- a/.storybook/manager.js
+++ b/.storybook/manager.js
@@ -1,0 +1,13 @@
+import { addons } from '@storybook/addons';
+import { create } from '@storybook/theming';
+
+addons.setConfig({
+    theme: create({
+        base: 'light',
+        /**
+         * NOTE: the font needs to be manually synchronized with the font in cloudify-ui-common
+         * since importing SCSS does not work in manager.js
+         */
+        fontBase: 'Roboto Regular'
+    })
+});
--- a/.storybook/manager.js
+++ b/.storybook/manager.js
@@ -1,6 +1,8 @@
 import { addons } from '@storybook/addons';
 import { create } from '@storybook/theming';
 
+import 'cloudify-ui-common/styles/font-Roboto-Regular.css';
+
 addons.setConfig({
     theme: create({
         base: 'light',


kubama:
Aren't the 'before' and 'after' images swapped?

qooban:
Is in the `after` snapshot a `Roboto` font? Does not look like it. See: https://www.dafont.com/roboto.font

Gelio:
@kubama @qooban As far as the devtools show, yeah, that's the Roboto font. I have updated the screenshots in the description, so take a look.

Since the sidebar outside of the `iframe` with the preview (the main content), maybe the font is not loaded there and there's some font isolation :thinking: so it falls back to some system font. I'll check it out

Gelio:
My presumption was right. The Roboto font itself was not loaded in the main document, and there was some fallback system one. I have updated the _After_ screenshot once again, now with the (hopefully) correct font

btw. that was the rendered font previously:
image

Gelio:
btw. I do not plan on releasing a new version of components, since it's nothing groundbreaking :smile:

.storybook/manager.js
```diff
@@ -0,0 +1,15 @@
...
+import { create } from '@storybook/theming';
+
+import 'cloudify-ui-common/styles/font-Roboto-Regular.css';
+
+addons.setConfig({
+    theme: create({
+        base: 'light',
+        /**
+         * NOTE: the font needs to be manually synchronized with the font in cloudify-ui-common
+         * since importing SCSS does not work in manager.js
```
qooban, line 11: I don't understand that comment. What does manually synchronized means?

qooban (commented):
Now the screenshot (`after`) looks like Roboto Regular 🙂

.storybook/manager.js
Gelio: If someone changes the font in `cloudify-ui-common`, they would have to manually change it in this file :slightly_smiling_face: 

This is different from the behavior in `preview.js` (the main frame of the documentation), where we reuse the font name as a variable exported from the theme (see the description)

Gelio: Let me know if you have a better wording for this comment that would be easier to understand :slightly_smiling_face:

kubama:
Out of curiosity, is the 'before' image really correct? If yes, then as for me the change in the font look is less than cosmetic. I wouldn't say the change is worth investing any time into.

.storybook/manager.js
qooban: No need to change 👍🏻

qooban:
@kubama I suspect that in the meantime (I reported that bug long time ago) Storybook changed theme and now we have sans serif font. Before it was for sure serif-like which looked odd comparing to the rest of the page which used sans serif like fonts (e.g. Roboto).

qooban (approved):

Gelio:
Great :slightly_smiling_face: I'm merging the PR, then

PR #74, opened 2021-07-05
RD-2690 Fix scrolling form to top when there are errors

Gelio:
When there are new errors in the form, the functionality for scrolling
the form back to the top to see them was broken since at least 2019-11.

The problem was that the DOM node for `this` cound not have been found
in a function component.

The fix is to use a ref to get a handle to the form and then invoke
scrolling on that.

There is also a new test that makes sure the scroll method is invoked.

## Video

I have installed this new version in Stage, here are the results:

https://user-images.githubusercontent.com/889383/124468262-2f67d200-dd99-11eb-99a3-69cd0a168600.mp4

## The plan
1. Merge this PR
2. Release a new **patch** version of `cloudify-ui-components`
3. Install this new version in `cloudify-stage` (I will skip `cloudify-blueprint-composer`)
4. Add a system test to make sure this scrolling works in Stage (tentative)
5. Raise a new PR for Stage based on points 3 and 4.

### DIFF
--- a/src/components/form/Form/Form.jsx
+++ b/src/components/form/Form/Form.jsx
@@ -1,9 +1,9 @@
-import React, { useEffect } from 'react';
+import React, { useEffect, useRef } from 'react';
 import ReactDOM from 'react-dom';
 import PropTypes from 'prop-types';
 import _ from 'lodash';
 
-import { Form as FormSemanticUiReact, Radio } from 'semantic-ui-react';
+import { Form as FormSemanticUiReact, Radio, Ref } from 'semantic-ui-react';
 
 import ErrorMessage from 'components/elements/ErrorMessage';
 import FormDropdown from 'components/elements/Dropdown';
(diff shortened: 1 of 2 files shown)


src/components/form/Form/Form.jsx
```diff
@@ -87,13 +87,10 @@ export default function Form({
...
 }) {
+    const formRef = useRef();
     useEffect(() => {
         if (scrollToError && !_.isEmpty(errors)) {
-            // eslint-disable-next-line react/no-find-dom-node
-            const formElement = ReactDOM.findDOMNode(this);
-            if (formElement) {
-                formElement.scrollIntoView({ behavior: 'smooth', block: 'start' });
-            }
+            formRef.current?.scrollIntoView?.({ behavior: 'smooth', block: 'start' });
```
Gelio, line 93: The optional chaining that checks whether `scrollIntoView` is available before calling it is for unit tests - it does not exist in jsdom and that results in some tests failing.

Alternatively, I can disable the `scrollToError` feature in the tests and leave this as

```suggestion
            formRef.current?.scrollIntoView({ behavior: 'smooth', block: 'start' });
```

alexmolev (approved):
LGTM

Gelio:
Thanks for a quick review :smile:

PR #76, opened 2021-07-26
RD-2789 Migrate some components to TS and use TS project references

Gelio:
This PR is a preparation for the actual code change of RD-2789 and it aims to solve some technical debt in the UI components repo.

It mostly consists of TS migrations of some components (particularly the `Form`, `Pagination`) which should help with typing in the Stage repo.

Moreover, I have used TS project references to hopefully speed up the TS compilation (it was fast before, but I guess using TS references does not hurt).

The aim of this PR is to have as few functional changes as possible.

I will release a minor version of cloudify-ui-components after merging this PR


.gitignore
```diff
@@ -7,3 +7,4 @@ es
 node_modules
 storybook-static
 typings
+*.tsbuildinfo
```
Gelio, line 10: `composite: true` results in this file being generated

babel.config.js
```diff
@@ -10,7 +10,8 @@ module.exports = api => {
                     root: ['./src']
                 }
             ],
-            testEnvironment && 'require-context-hook'
+            testEnvironment && 'require-context-hook',
+            '@babel/proposal-class-properties'
```
Gelio, line 14: Necessary to allow using any class properties, `static` or not

src/components/elements/ErrorMessage/ErrorMessage.tsx
```diff
@@ -0,0 +1,110 @@
+import React, { ComponentProps, ReactElement, useEffect, useState } from 'react';
```
Gelio, line 1: Unfortunately, I've made too many changes and git does not detect the rename :worried:

src/components/elements/ErrorMessage/ErrorMessage.tsx
```diff
@@ -0,0 +1,110 @@
...
+        return () => {
+            clearTimeout(messageVisibilityTimeout);
+        };
+    }, [error]);
+
+    return !isEmpty(error) ? (
+        <Message error className={className} hidden={hidden} onDismiss={handleDismiss}>
+            {/* NOTE: the error is not empty, thus, is defined */}
+            {/* eslint-disable-next-line @typescript-eslint/no-non-null-assertion */}
+            {renderErrorMessage(error!, header)}
```
Gelio, line 66: I have extracted the `if`s that we previously had before the `return` to a separate function to hopefully make it more clear

src/components/form/Form/Form.tsx
```diff
@@ -81,13 +89,12 @@ import './Form.css';
 export default function Form({
     children,
     errors,
-    errorMessageHeader,
-    onErrorsDismiss,
-    onSubmit,
```
Gelio, line 86: `onSubmit` is a part of `formProps` now (as it's a part of `StrictFormProps`)

Gelio:
I have decided to revert changes to `ErrorMessage` because there were too many errros in stage because `useErrors` is used as a bucket for multiple types of errors instead of having errors of just a single type :worried:

src/components/data/common/Pagination/Pagination.tsx
```diff
@@ -22,11 +53,11 @@ export default class Pagination extends Component {
         this.changePageSize = this.changePageSize.bind(this);
     }
 
-    componentDidUpdate(prevProps) {
+    componentDidUpdate(prevProps: PaginationProps): void {
         const { pageSize, totalSize } = this.props;
         const { currentPage } = this.state;
 
-        const changedProps = {};
+        const changedProps: { pageSize?: number; currentPage?: number } = {};
```
qooban, line 60: Just wondering - is there a better way of defining that type? (reusing `PaginationProps` somehow or extracting relevant part from it). 

I'm afraid it may overcomplicate the solution, so just asking.

src/components/data/common/Pagination/Pagination.tsx
```diff
@@ -1,16 +1,47 @@
...
+        pageSize: PropTypes.number,
+        sizeMultiplier: PropTypes.number
+    };
+
+    // eslint-disable-next-line react/static-property-placement
+    static defaultProps = {
+        totalSize: 0,
+        pageSize: Pagination.PAGE_SIZE_LIST(5)[0],
+        sizeMultiplier: 5
+    };
```
qooban, line 41: I assume that we are keeping propTypes definition only due to the fact that `DataTable` is not yet migrated to TS, right?

src/components/elements/ErrorMessage/ErrorMessage.jsx
```diff
@@ -49,14 +49,12 @@ export default function ErrorMessage(props) {
...
 
-    return (
-        !_.isEmpty(error) && (
-            <Message error className={className} hidden={hidden} onDismiss={handleDismiss}>
-                <Message.Header>{errorHeader}</Message.Header>
-                {_.isArray(error) ? <Message.List items={error} /> : <Message.Content>{errorMessage}</Message.Content>}
-            </Message>
-        )
-    );
+    return !_.isEmpty(error) ? (
```
qooban, line 52: Optional:

Maybe import `isEmpty` and `isArray` directly from `lodash` at once? I see there are 2 more in this file - `isObject` and `isEqual`.

src/components/form/Form/Form.tsx
```diff
@@ -1,12 +1,11 @@
-import React, { useEffect, useRef } from 'react';
```
qooban, line 1: Great that you migrated that file. I think it means that you did part of RD-1837 🙂 

Optional:

Maybe when you'll be doing changes in cloudify-stage you can just update relevant places (remove TODO and use `Form` component directly)?

src/components/data/common/Pagination/PaginationInfo.tsx
```diff
@@ -1,16 +1,27 @@
...
 
-/**
- * @return {null}
- */
+interface PaginationInfoProps {
+    pageSize: number;
+    onPageSizeChange: (pageSize: string) => void;
+    totalSize: number;
+    currentPage: number;
+    sizeMultiplier: number;
```
qooban, line 12: As they are optional, shouldn't we define them this way:
```suggestion
    currentPage?: number;
    sizeMultiplier?: number;
```
?

Gelio:
Looks like there is some work to be done before this is reviewable. First of all, I need to fix the CI (there are some storybook failures) and make sure it all works when used in Stage.

src/components/data/common/Pagination/Pagination.tsx
```diff
@@ -1,16 +1,47 @@
...
 import './Pagination.css';
 
-export default class Pagination extends Component {
-    constructor(props, context) {
-        super(props, context);
+interface PaginationProps {
+    fetchData: () => void;
+    totalSize: number;
+    pageSize: number;
+    sizeMultiplier: number;
```
qooban, line 15: Since they are optional shouldn't we define them this way:
```suggestion
    totalSize?: number;
    pageSize?: number;
    sizeMultiplier?: number;
```
?

src/components/data/common/Pagination/Pagination.tsx
```diff
@@ -127,19 +159,3 @@ export default class Pagination extends Component {
...
-    totalSize: PropTypes.number,
-    pageSize: PropTypes.number,
-    sizeMultiplier: PropTypes.number
-};
-
-Pagination.defaultProps = {
-    totalSize: 0,
-    pageSize: Pagination.PAGE_SIZE_LIST(5)[0],
-    sizeMultiplier: 5
-};
```
Gelio, line 145: Moved to `static` class properties

src/components/elements/ErrorMessage/ErrorMessage.jsx
```diff
@@ -49,14 +49,12 @@ export default function ErrorMessage(props) {
...
+        <Message error className={className} hidden={hidden} onDismiss={handleDismiss}>
+            <Message.Header>{errorHeader}</Message.Header>
+            {_.isArray(error) ? <Message.List items={error} /> : <Message.Content>{errorMessage}</Message.Content>}
+        </Message>
+    ) : null;
```
Gelio, line 57: This missing `null` prevented using this component in `Stage` directly

src/components/form/Form/Form.tsx
```diff
@@ -21,6 +20,15 @@ import UrlOrFileInput from '../UrlOrFileInput';
 
 import './Form.css';
 
+interface FormProps extends Omit<StrictFormProps, 'error'> {
+    errors?: unknown;
```
qooban, line 24: Did you intentionally used `unknown`? 

We know that type. In the PropTypes language it is `ErrorMessage.propTypes.error` which equals to:
```
  PropTypes.oneOfType([
        PropTypes.string,
        PropTypes.array,
        PropTypes.shape({ header: PropTypes.string, message: PropTypes.string })
    ])
```

Or maybe it is not possible to define such weird type in TS? 🙂

src/components/form/Form/Form.tsx
```diff
@@ -104,15 +111,16 @@ export default function Form({
     return (
         <Ref innerRef={formRef}>
             {/* eslint-disable-next-line react/jsx-props-no-spreading */}
-            <FormSemanticUiReact {...formProps} onSubmit={onSubmit} error={!_.isEmpty(errors)}>
-                <ErrorMessage header={errorMessageHeader} error={formattedErrors} onDismiss={onErrorsDismiss} />
+            <FormSemanticUiReact {...formProps} error={!_.isEmpty(errors)}>
+                {/* eslint-disable-next-line @typescript-eslint/no-explicit-any */}
+                <ErrorMessage header={errorMessageHeader} error={formattedErrors as any} onDismiss={onErrorsDismiss} />
```
qooban, line 116: If you manage to change `errors` type, then we probably can avoid using `any` here.

qooban (commented):
Generally great work migrating to TS 🙂 

I added some comments, mostly minor. In some cases I just probably need some clarification.

qooban:
> Looks like there is some work to be done before this is reviewable. First of all, I need to fix the CI (there are some storybook failures) and make sure it all works when used in Stage.

OK 🙂 I did the review anyway as I noticed this comment too late 🙂

> I'm already frustrated enough as I have spent 4h trying to fix Stage after all the TS migrations

Yeah... I believe it might be painful.

Gelio:
So far I have only rebased the branch on top of `master` without doing any code changes. Conflicts were resolved automatically by git. I will be adding new commits from now on

src/components/data/common/Pagination/Pagination.tsx
Gelio: I don't think it matters much, as it's only an object that is used locally here to make the computation easier. Its individual properties are then used separately in https://github.com/cloudify-cosmo/cloudify-ui-components/blob/45fef87eabea093610a197fc3edac43ecb2ebaeb/src/components/data/common/Pagination/Pagination.tsx#L74

Reusing `PaginationProps` would mean:

```
Partial<Pick<PaginationProps, 'pageSize' | 'currentPage'>>
```

which IMO is more complex and does not yield many benefits :smile: 

I would rather leave it unchanged :slightly_smiling_face:

src/components/data/common/Pagination/Pagination.tsx
Gelio: Yup, exactly :slightly_smiling_face:

src/components/elements/ErrorMessage/ErrorMessage.jsx
Gelio: Good idea, let me change it

src/components/form/Form/Form.tsx
Gelio: >  I think it means that you did part of RD-1837 slightly_smiling_face

Yup :smile: 

> Maybe when you'll be doing changes in cloudify-stage you can just update relevant places (remove TODO and use Form component directly)?

That was my intention :smile: I have already removed `UnsafelyTypedForm` from `cloudify-stage` locally, I will push it out in a PR for RD-2789

src/components/data/common/Pagination/PaginationInfo.tsx
Gelio: That's the magic of TS and `defaultProps`. The short answer is: they should be required and TypeScript does some magic because the properties are optional when you are trying to use the component, as they appear in `defaultProps`

Proof:


image


https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/default_props/ for some more information on the topic (but it does not describe it too well)

src/components/data/common/Pagination/Pagination.tsx
Gelio: The answer is the same as for https://github.com/cloudify-cosmo/cloudify-ui-components/pull/76#discussion_r677256691 :slightly_smiling_face:

src/components/form/Form/Form.tsx
Gelio: Defining such a type is possible in TypeScript (it should be a union type using `|`). I did it this way initially when I was migrating `ErrorMessage` to TS (because this `errors` prop is passed to `ErrorMessage`), but it turned out that this in actuality the `errors` prop takes many shapes and sizes in `stage` and it's a lot more work to solve all the TS problems in stage and come up with a type that works here :smile: 

Thus, I decided to use `unknown` to make it simple

src/components/elements/ErrorMessage/ErrorMessage.jsx
Gelio: Done

Gelio:
Looks like I managed to fix the issues with storybook. It turns out that `build-storybook` does not play well with static properties on component classes :slightly_frowning_face: Running `npm run storybook` worked fine. Weird. The bug was very cryptic ("unexpected token" but the code looked fine, even ES5-compliant). I could not find any other solution online, so I did a dirty fix in https://github.com/cloudify-cosmo/cloudify-ui-components/pull/76/commits/533ccfe8e44092974a2317e2700286ce6b6499ea

Besides, looks like the PR is reviewable now and the CI passed. I have built it locally and used it in Stage - it looks to be working well and there are some comments/warnings we can remove, which I'm working on right now.

After the PR is merged I plan on releasing a minor version of `cloudify-ui-components` and work on using it in Stage :slightly_smiling_face:

src/components/data/common/Pagination/PaginationInfo.tsx
qooban: OK, clear 👍🏻 Thanks for explanation.

src/components/data/common/Pagination/Pagination.tsx
qooban: 👍🏻

src/components/form/Form/Form.tsx
qooban: Got it. It's understandable, then.

src/components/form/Form/Form.tsx
qooban: Great 🙂

src/components/form/Form/Form.tsx
qooban: I believe so. Thanks for the conversion attempt.

src/components/data/common/Pagination/Pagination.tsx
qooban: Let's leave it unchanged.

qooban (approved):
Looks good 👍🏻 Thanks for the clarifications.

Gelio:
@stepek do you plan on reviewing this PR? :slightly_smiling_face:

stepek (approved):
LGMT!

Gelio:
Thanks for reviews! :smile: 

I'll merge the PR and raise a minor version of the UI components

PR #77, opened 2021-07-26
RD-2789 Take the current page size into account when showing pagination

Gelio:
Include the current page size into account when deciding whether to show
the pagination or not.

This fixes a problem with the pagination not being shown when the
`pageSize` was smaller than `totalSize`, but the `sizeMultiplier` was
greater than `totalSize` (which was the main criterium when deciding
to show the pagination).

## Screenshot

When used in Stage:

image


qooban (approved):
LGTM 👍🏻

Gelio:
I'll merge the PR with just a single approval, as the change seems small and is validated by a unit test

Thanks @qooban for a quick review :+1:

PR #79, opened 2021-08-10
RD-2826 Make `KeyIndicator` component more customizable

qooban:
Main change introduced in this PR is an option to provide URL for the image in `KeyIndicator` component and make all props optional. Along with that change I decided to migrate `KeyIndicator` component to TS and add small ESLint override.

This change is meant to support implementation of those tickets: RD-2826, RD-2930.

### Screenshot
* For customized "Number of Deployments" widget:
  image
* For new "Number of clouds" widget:
  image


stepek (approved):
LGTM

kubama (approved):
Looks good 👍.

I have another PR for common-components so maybe we could release both changes at once.

qooban:
Thanks for quick reviews and approvals. I'll merge this PR, then.

> I have another PR for common-components so maybe we could release both changes at once.
@kubama Yes, makes sense to release just one version (probably minor) and that way we will just need to do one change in docs.cloudify.co.

PR #78, opened 2021-08-10
RD-1151 Added support for `className` and `style` props where missing

kubama:
Although there are 35 modified files this PR should be relatively easy to review as there is just a single change (`className` and `style` props introduction) applied across all the files.

PR #81, opened 2021-08-12
RD-1151 Migrated some components to TS

kubama:
Changes introduced in https://github.com/cloudify-cosmo/cloudify-ui-components/pull/78 caused PropTypes in three components to be ignored by TS. As a result optional properties in these components were no longer treated as such. Migrating these components to TS solves the issue.

NOTE: one minor type fix in stage will be necessary when upgrading.


qooban (approved):
Looks good! 👍🏻 

Another step towards full TS migration done 🙂 
Once you release new version of cloudify-ui-components I can update https://github.com/cloudify-cosmo/docs.getcloudify.org/pull/1911 .
Are you planning to create a PR in stage also to do that minor type fix?

PR #85, opened 2021-09-09
RD-3100 Extended `HighlightText` component to accept all props supported by underlying `react-syntax-highlighter`

kubama:
## Description
In order to fix RD-3100 some customization to `HighlightText` is needed that is currently not supported.

## Screenshots / Videos
N/A

## Checklist

- [X] My code follows the UI Code Style.
- [X] I verified if that change requires a documentation update.

## Tests
N/A

## Documentation
N/A


src/components/data/HighlightText/HighlightText.tsx
```diff
@@ -25,33 +25,26 @@ SyntaxHighlighter.registerLanguage('yaml', yaml);
  * * JSON
  * * Python
  * * YAML
+ *
+ * All props supported by `react-syntax-highlighter` can be specified.
  */
-export default function HighlightText(props) {
-    const { children, language } = props;
-
+const HighlightText: FunctionComponent<SyntaxHighlighterProps> = ({
```
qooban, line 31: Having external `SyntaxHighlighterProps` how does the table with props in the documentation looks like?

I'm referring to: https://docs.cloudify.co/ui-components/2.7.1/?path=/docs/data-highlighttext--basic which has right now:
Screenshot from 2021-09-09 15-22-22

How will new storybook page for `HighlightText` look like?

src/components/data/HighlightText/HighlightText.tsx
```diff
@@ -25,33 +25,26 @@ SyntaxHighlighter.registerLanguage('yaml', yaml);
...
+    children = '',
+    codeTagProps,
+    language = 'json',
```
qooban, line 34: As we register only predefined set of languages (see `SyntaxHighlighter.registerLanguage` calls), shouldn't we allow only that set of languages?

qooban (commented):
Commented.

src/components/data/HighlightText/HighlightText.tsx
kubama: We can always register additional languages, so I would say no.

src/components/data/HighlightText/HighlightText.tsx
kubama: The new docs will not have the props listed. But the doc mentions which props are supported (I can add direct link).
This is something we already have in storybook, for example here: https://docs.cloudify.co/ui-components/2.7.1/?path=/docs/elements-checkmark--basic. The doc mentions that the component supports `Icon` props but does not list them. I don't see any issue with such approach.

qooban: As `SyntaxHighlighter` does not come from Semantic-UI-React, I think that having a link here is necessary. Thanks for adding it.

src/components/data/HighlightText/HighlightText.tsx
qooban: Hmm... How widget developer would know about it?
I think we should either add a note about it to the description or allow just predefined set.

qooban (commented):
Answered comments.

src/components/data/HighlightText/HighlightText.tsx
qooban: Also, widget developer would need to import `react-syntax-highlighter` to call `registerLanguage` as we don't expose it through Stage API.

kubama: OK, I brought back limited set of values for this prop

qooban (approved):

PR #116, opened 2022-01-26
RD-1837: Migration to typescript. FormField and a few Form related components.

jawamartus:
## Description

The PR provides a migration to typescript for FormField (RD-1837) and some other related components.

I didn't manage to force resolver to use absolute paths. That's why I continue to use relative paths for typescript files.
## Checklist

- [x] My code follows the UI Code Style.
- [x] I verified if that change requires a documentation update.


src/components/form/Form/FormField.tsx
```diff
@@ -0,0 +1,48 @@
+import React, { memo } from 'react';
+import _ from 'lodash';
+
+import { Form, StrictFormFieldProps } from 'semantic-ui-react';
```
qooban, line 4: Probably more correct would be to have:
```suggestion
import { Form } from 'semantic-ui-react';
import type { StrictFormFieldProps } from 'semantic-ui-react';
```

src/components/form/Form/FormField.tsx
```diff
@@ -0,0 +1,48 @@
...
+    help?: React.ReactNode;
+}
+
+/**
+ * `Form.Field` is a component to present field and is used in Form component.
+ *
+ * `Form.Field` is a wrapper for [Semantic UI-React's Form.Field component](https://react.semantic-ui.com/collections/form),
+ * so all properties of that component can be used here.
+ */
+export function FormFieldWithoutMemo({
```
qooban, line 25: In `cloudify-ui-components` as well as in `cloudify-stage` we use the following convention:
```
import type { FunctionComponent } from 'react';
const ComponentName: FunctionComponent<ComponentNameProps> = ({ <props list> }) => {
```
For consistency, maybe worth to use it here as well?

`FunctionComponent` already has `children` prop defined 🙂

qooban, line 25: There's `export default FormField;`, so is it necessary to export `FormFieldWithoutMemo`?

src/components/popups/PopupHelp/PopupHelp.tsx
```diff
@@ -1,17 +1,35 @@
...
+
 /**
  * PopupHelp is a component which uses [Popup](https://react.semantic-ui.com/modules/popup) component to display
  * help popup.
  *
  * All props supported by the underlaying `Popup` component are passed down to it.
  */
-export default function PopupHelp(props) {
-    const { header, content, trigger, ...popupProps } = props;
+export default function PopupHelp(props: Props) {
```
qooban, line 31: Let's use already existing convention for defining TS components (see my other comment).

qooban (commented):
Added some comments. 

It's a pity you didn't manage to solve import resolver issue, but for now I think it's fine to keep relative path imports.

I've got few general questions:
1. Will the change you did here be sufficient to remove both - `UnsafelyTypedFormField` and `UnsafelyTypedFormGroup` from `cloudify-stage`?
2. Will the change be sufficient for using in TS files other `Form` sub-components?
There is quite a lot:
```
Form.Button = FormSemanticUiReact.Button;
Form.Checkbox = Checkbox;
Form.ColorPicker = ColorPicker;
Form.DatePicker = DatePicker;
Form.Date = DateInput;
Form.DateRange = DateRangeInput;
Form.Divider = FormDivider;
Form.Dropdown = FormDropdown;
Form.Field = FormField;
Form.File = FileInput;
Form.Group = FormSemanticUiReact.Group;
Form.Input = FormSemanticUiReact.Input;
Form.Json = JsonInput;
Form.Radio = Radio;
Form.TextArea = FormSemanticUiReact.TextArea;
Form.UrlOrFile = UrlOrFileInput;
Form.Time = TimeInput;
```
3. There's an issue in CI pipeline with `npm run build-storybook`. Have you tried to figure out what's the problem?

src/components/form/Form/FormField.tsx
jawamartus: We are using the FormFieldWithoutMemo inside of storybook file. It was previously here before rewrite. I guess it was intentional.

jawamartus: In Form.Field component we require children prop. FunctionComponent type it as optional.

jawamartus:
> 1. Will the change you did here be sufficient to remove both - `UnsafelyTypedFormField` and `UnsafelyTypedFormGroup` from `cloudify-stage`?

Yes, I checked it on the code in this Draft PR: https://github.com/cloudify-cosmo/cloudify-stage/pull/1798

> 
>     2. Will the change be sufficient for using in TS files other `Form` sub-components?
>        There is quite a lot:
> (...)
I don't know the answer. Is that in the scope of this ticket to rewrite all of them? I would limit this PR to not all of the components aforementioned.

> ```
> 
>     3. There's an issue in CI pipeline with `npm run build-storybook`. Have you tried to figure out what's the problem?
It's solved!

src/components/popups/PopupHelp/PopupHelp.tsx
jawamartus: Would do you mean by existing convention? Do you mean Props type name? I will change it.

I've done it similar to the component: `src/components/layout/ProductVersion/ProductVersion.tsx`.

src/components/form/Form/Form.tsx
```diff
@@ -101,7 +122,7 @@ export default function Form({
         }
     }, [errors]);
 
-    let formattedErrors = errors;
+    let formattedErrors: any = errors;
```
kubama, line 125: I think we could be more specific here regarding the type as we know the type of `errors`.

src/components/form/Form/Form.tsx
```diff
@@ -86,7 +107,7 @@ interface FormProps extends Omit<StrictFormProps, 'error'> {
  *
  */
 
-export default function Form({
+function Form({
```
kubama, line 110: Is it any better to export in a separate line as opposed to the place of declaration?

src/components/form/Form/FormDivider.tsx
```diff
@@ -1,8 +1,18 @@
 import React from 'react';
-import PropTypes from 'prop-types';
-
+import type { ReactNode } from 'react';
 import { Header } from 'semantic-ui-react';
 
+type FormDividerProps = {
```
kubama, line 5: This component passes down props to the underlaying `Header` component - if so, shouldn't the type for its props extend `HeaderProps`? Or possibly `HeaderProps` could be used directly in place if this type.

src/components/form/Form/FormDivider.tsx
```diff
@@ -11,25 +21,10 @@ import { Header } from 'semantic-ui-react';
  *
  * Accessible as `Form.Divider`.
  */
-export default function FormDivider({ children, className, ...headerProps }) {
+export default function FormDivider({ children, className = '', ...headerProps }: FormDividerProps) {
```
kubama, line 24: I suspect the default empty string for `className` is not necessary.

src/components/popups/PopupHelp/PopupHelp.tsx
```diff
@@ -1,17 +1,35 @@
 import React from 'react';
-import PropTypes from 'prop-types';
 import _ from 'lodash';
 import { Icon } from 'semantic-ui-react';
 import Popup from '../Popup';
 
+interface PopupHelpProps {
```
kubama, line 6: This type should somehow extend `PopupProps` as the related component passes props down to `Popup`.

src/components/form/GenericField/GenericField.jsx
```diff
@@ -6,7 +6,7 @@ import { Input } from 'semantic-ui-react';
 import Dropdown from 'components/elements/Dropdown';
 import ErrorMessage from 'components/elements/ErrorMessage';
 import Checkbox from 'components/form/Checkbox';
-import FormField from 'components/form/Form/FormField';
+import FormField from '../Form/FormField';
```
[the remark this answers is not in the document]
jawamartus: The import resolver returns errors for importing tsx files by absolute path.

src/components/form/Form/Form.tsx
jawamartus: Yes. Using default export here causes some problems with storybook-build command for tsx files.

src/components/form/Form/FormDivider.tsx
jawamartus: It was previously here set by defaultProps. I'm afraid that it could return error in some cases if I change it.

src/components/form/Form/Form.tsx
jawamartus: The data type is very complex here. However I made it more specific.

src/components/form/Form/FormDivider.tsx
jawamartus: Fixed.

src/components/popups/PopupHelp/PopupHelp.tsx
jawamartus: Thanks! I changed it.

src/components/form/GenericField/GenericField.jsx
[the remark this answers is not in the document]
kubama: I guess we can consider it to be out of scope of this PR, but did look into possibility of fixing the resolver? I think it should work the same way for both JS and TS files.

src/components/form/Form/Form.tsx
```diff
@@ -86,33 +108,35 @@ interface FormProps extends Omit<StrictFormProps, 'error'> {
...
+}: FormProps): ReactElement | null {
     const formRef = useRef<HTMLElement>(null);
     useEffect(() => {
         if (scrollToError && !_.isEmpty(errors)) {
             formRef.current?.scrollIntoView?.({ behavior: 'smooth', block: 'start' });
         }
     }, [errors]);
 
-    let formattedErrors = errors;
+    let formattedErrors: undefined | Array<string> | ReactElement<unknown, string | JSXElementConstructor<any>>;
```
kubama, line 126:
```suggestion
    let formattedErrors: undefined | Array<string> | JSX.Element;
```
Would be shorter and in line with `errors` type.

src/components/form/Form/FormDivider.tsx
```diff
@@ -1,7 +1,18 @@
 import React from 'react';
-import PropTypes from 'prop-types';
-
+import type { ReactNode, FunctionComponent } from 'react';
 import { Header } from 'semantic-ui-react';
+import type { HeaderProps } from 'semantic-ui-react';
+
+export interface FormDividerProps extends HeaderProps {
```
kubama, line 6: I think `HeaderProps` already include `children` and `className`, so maybe `HeaderProps` could be used directly in place of `FormDividerProps`?

src/components/form/Form/FormField.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { Form } from 'semantic-ui-react';
+import type { StrictFormFieldProps } from 'semantic-ui-react';
+import PopupHelp from '../../popups/PopupHelp';
+import FieldLabel from '../FieldLabel/FieldLabel';
+
+export interface FormFieldProps extends StrictFormFieldProps {
+    /**
+     * wrapped form fields
+     */
+    children: React.ReactNode;
```
kubama, line 14: Isn't `children` already part of `StrictFormFieldProps`?

src/components/popups/PopupHelp/PopupHelp.tsx
```diff
@@ -0,0 +1,46 @@
...
+    header?: React.ReactNode;
+    /**
+     * help popup triggering element (on hover and focus)
+     */
+    trigger?: React.ReactNode;
+    /**
+     * force Popup open
+     */
+    open?: boolean;
+}
```
kubama, line 25: Looks like `content`, `header` and `trigger` are already part of `PopupProps`.

jawamartus: Somehow yes, but we don't pass content, header and trigger inside of popupProps object to Popup component. We use them inside of PopupHelp.

And they are different. PopupProps.content is typed as optional. But it's required to run PopupHelp.

What is more the type declaration for props used directly in PopupHelp component should be independent from Popup component typing. It makes PopupHelp less coupled with Popup component. Please consider what would happen with PopupHelp typing if the property 'content' or 'trigger' would be removed from PopupProps in the next version of semantic-ui-react.

src/components/form/Form/FormDivider.tsx
jawamartus: I prefer to keep `className` here for similar reasons mentioned in one of my previous comments.

About children property - it's required by our component and not required by HeaderProps typing. Our component have more strict needs.

src/components/popups/PopupHelp/PopupHelp.tsx
kubama: > Somehow yes, but we don't pass content, header and trigger inside of popupProps object to Popup component.

Actually we do. `<Popup.Header>`, `<Popup.Trigger>` and `<Popup.Content>` can be actually passed directly as props on `Popup` element.

> PopupProps.content is typed as optional. But it's required to run PopupHelp.

It was originally marked as required in `propTypes`, but there is no reason for it to be so as the underlaying `Popup` can work perfectly without it. I think we can relax this.

> What is more the type declaration for props used directly in PopupHelp component should be independent from Popup component typing. It makes PopupHelp less coupled with Popup component.

I think the opposite 🙂. `PopupHelp` implementation is strictly coupled with the underlaying `Popup` component, so it should be the same with props typing. If anything changes in `Popup` props typing then `PopupHelp` props typing should reflect this change as well. For example, if in future SUI release `content` type was restricted (lets say it supported `string` only but no more `JSX.Element`) then `content` prop type in `PopupHelp` should reflect this change - if not then we would not catch an erroneous code that could be still passing `JSX.Element` to it.

> Please consider what would happen with PopupHelp typing if the property 'content' or 'trigger' would be removed from PopupProps in the next version of semantic-ui-react.

The whole component would then break due to the implementation, it has nothing to do with props typing.

src/components/form/Form/FormField.tsx
kubama: There is no reason to require this prop if the component works correctly without it. I assume it does as it's not required for `Form.Field`. We can relax this restriction as it doesn't break anything.

src/components/form/Form/FormDivider.tsx
kubama: My answer here is similar as in other comments 🙂.

src/components/form/GenericField/GenericField.jsx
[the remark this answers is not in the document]
jawamartus: I was trying to fix it without result yet. It's complex - the issue concern storybook, tests and eslint as well. All of them seems to return different errors in different cases when I try to use absolute path for tsx inside of jsx files.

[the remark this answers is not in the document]
jawamartus: I can try to solve it again and create separate PR for it if I would succeed.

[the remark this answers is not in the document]
kubama: Ok, lets leave it for a separate ticket then.

src/components/form/Form/FormDivider.tsx
jawamartus: I refactored the component a little. Thanks for the valuable discussion!

src/components/form/Form/FormField.tsx
jawamartus: I refactored the component a little. Thanks for the valuable discussion!

src/components/popups/PopupHelp/PopupHelp.tsx
jawamartus: I refactored the component a little. Thanks for the valuable discussion!

qooban (commented):
Like it! Types reusing makes this change really cool 👍🏻
I don't have any more comments than @kubama .

> I don't know the answer. Is that in the scope of this ticket to rewrite all of them? I would limit > this PR to not all of the components aforementioned.

Agree, let's limit this PR and the ticket to the changes you already did and make sure that we can use `Form.Field`, `Form.Group` instead of artificially created `UnsafelyTyped...` components in Stage.

src/components/form/Form/Form.tsx
```diff
@@ -86,33 +109,35 @@ interface FormProps extends Omit<StrictFormProps, 'error'> {
...
     } else if (_.isObject(errors) && !React.isValidElement(errors)) {
         formattedErrors = _.valuesIn(errors);
+    } else {
+        formattedErrors = errors;
     }
 
     return (
         <Ref innerRef={formRef}>
             <FormSemanticUiReact {...formProps} error={!_.isEmpty(errors)}>
                 {/* eslint-disable-next-line @typescript-eslint/no-explicit-any */}
```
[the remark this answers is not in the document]
jawamartus: Removed.

src/components/form/Form/Form.tsx
```diff
@@ -21,9 +22,31 @@ import UrlOrFileInput from '../UrlOrFileInput';
...
-    errors?: unknown;
+    /**
+     * primary content
+     */
+    children: React.ReactNode;
+
+    /**
+     * function called on form submission
+     */
+    onSubmit?: () => void;
```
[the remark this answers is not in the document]
jawamartus: Removed.

src/components/form/Form/FormDivider.tsx
jawamartus: Tested and changed.

PR #117, opened 2022-02-09
RD-4208: Raised cloudify-ui-common version and other package update

jawamartus:
https://cloudifysource.atlassian.net/browse/RD-4208

## Description

In the scope of the ticket I rised version of cloudify-common package and after discussion with Jakub N. I also rised version of other not up to date packages.

## Checklist

- [ ] My code follows the UI Code Style.
- [ ] I verified if that change requires a documentation update.


jest.config.js
```diff
@@ -4,6 +4,7 @@ const path = require('path');
 const stubLink = path.resolve(__dirname, '.jest/stub.js');
 
 module.exports = {
+    testEnvironment: 'jsdom',
```
jawamartus, line 7: In the previous version of jest the default testEnvironment was "jsdom". After the upgrade of major version, now it is "node". That's why this configuration option is now present and set explicitly.

src/components/elements/EditableLabel/EditableLabel.jsx
```diff
@@ -40,7 +40,7 @@ export default function EditableLabel({
 
         setEditing(false);
         onChange(currentValue);
-    }
+    }, [currentValue, value, setEditing, onCancel, invalidValues, setError, onError, onChange]);
```
qooban, line 43: I'm wondering if it is necessary to put here `setEditing`, `onCancel`, `setError`, `onError` and `onChange`.
In `src/components/form/UrlOrFileInput/UrlOrFileInput.jsx` there are no deps set in the hooks.

qooban (commented):
Great that you've managed to make `audit` green again! 👍🏻 
Added just one comment for clarification.

Some general comments:

1. RD-4208 is a sub-task of RD-2570, but is it actually necessary to change anything in `cloudify-ui-components` in order to implement RD-2570? If not let's take it out from RD-4208 to become independent task or consult with @leon-good-life and move it to RD-3222.  

2. I think that it might be a good idea to do quick test on changed components in stage (once you release new version of `cloudify-ui-components`)

jawamartus:
>     1. [RD-4208](https://cloudifysource.atlassian.net/browse/RD-4208) is a sub-task of [RD-2570](https://cloudifysource.atlassian.net/browse/RD-2570), but is it actually necessary to change anything in `cloudify-ui-components` in order to implement [RD-2570](https://cloudifysource.atlassian.net/browse/RD-2570)? If not let's take it out from [RD-4208](https://cloudifysource.atlassian.net/browse/RD-4208) to become independent task or consult with @leon-good-life and move it to [RD-3222](https://cloudifysource.atlassian.net/browse/RD-3222).
I done! I make it by mistake. I thought that we run some cypress tests inside of components.

Do I understand correctly that we run cypress tests inside of composer repository? If yes, it looks like I need to raise the ui-common package there to close [RD-2570](https://cloudifysource.atlassian.net/browse/RD-2570)

> 
>     2. I think that it might be a good idea to do quick test on changed components in stage (once you release new version of `cloudify-ui-components`

I will do that.

src/components/elements/EditableLabel/EditableLabel.jsx
jawamartus: About setEditing and setError you are right for sure (yellow table with text) https://reactjs.org/docs/hooks-reference.html#usestate

However I would keep other function references there. I don't want to limit the component users to pass only one version of function in props on the beginning.

I made a few changes to make it more bulletproof and correct.
https://github.com/cloudify-cosmo/cloudify-ui-components/pull/117/commits/6d545f34e4ef5250d69f97bda98d825d021e7bbe

qooban (approved):
> Do I understand correctly that we run cypress tests inside of composer repository? If yes, it looks like I need to raise the ui-common package there to close RD-2570

Yes, exactly. It's needed in both - Stage and Composer.

PR #118, opened 2022-02-11
RD-1837: Upgrading typescript types for Form component

jawamartus:
https://cloudifysource.atlassian.net/browse/RD-1837
## Description

Small amendment before release of a new version of the component package. It makes cloudify-stage working after Form component migration to typescript.

## Checklist

- [x] My code follows the UI Code Style.
- [x] I verified if that change requires a documentation update.


src/components/form/Form/Form.tsx
```diff
@@ -25,7 +25,7 @@ interface FormProps extends Omit<StrictFormProps, 'error'> {
     /**
      * string with error message or object with fields error messages (syntax described above)
```
qooban, line 26: That's not accurate right now. 

The description is below not above 🙂 (L88)
I'm not sure, but maybe worth to move it here?

src/components/form/Form/Form.tsx
```diff
@@ -25,7 +25,7 @@ interface FormProps extends Omit<StrictFormProps, 'error'> {
...
      */
-    errors?: string | [] | JSX.Element | { header: string; message: string };
+    errors?: string | Record<string, any> | JSX.Element;
```
qooban, line 28: Just a question - I assume that `[]` was added here by mistake and there's no use-case for that, am I right?

qooban (commented):
Good that you detected the issue before releasing new version 👍🏻 

Just a minor comment added.

src/components/form/Form/Form.tsx
jawamartus: Fixed. Also I improved the type a little bit.

src/components/form/Form/Form.tsx
```diff
@@ -23,9 +23,9 @@ import './Form.css';
 
 interface FormProps extends Omit<StrictFormProps, 'error'> {
     /**
-     * string with error message or object with fields error messages (syntax described above)
+     * string with error message or object with fields error messages (syntax described below)
      */
-    errors?: string | [] | JSX.Element | { header: string; message: string };
+    errors?: string | Record<string, string>;
```
qooban, line 28: Hm... I checked it deeper, and it seems like it is still not fine.
I think we should keep `JSX.Element` here. 

You can check repo's history. Leon added support for component-like errors recently (RD-1607).

src/components/form/Form/Form.tsx
```diff
@@ -23,9 +23,9 @@ import './Form.css';
 
 interface FormProps extends Omit<StrictFormProps, 'error'> {
     /**
-     * string with error message or object with fields error messages (syntax described above)
+     * string with error message or object with fields error messages (syntax described below)
```
qooban, line 26: I'm wondering wouldn't it be misleading. There are two cases when this text is displayed:
1. In code - when you are reading the code directly
2. In Storybook - when you are browsing through https://docs.cloudify.co/ui-components/2.11.0/?path=/docs/form-form--basic

In the first case it's fine, but in the second no... So probably worth to do a proper fix and move "Errors" section (L88) here. WDYT?

src/components/form/Form/Form.tsx
jawamartus: Thanks for pointing it out. I reverted JSX.Element.

src/components/form/Form/Form.tsx
jawamartus: You are right. Moved!

PR #125, opened 2022-02-25
RD-2792: GenericField - improving typing and refactoring

jawamartus:
[Build Status](https://jenkins.cloudify.co/job/Stage-UI-System-Test/2173/)
[Build Status](https://jenkins.cloudify.co/job/Stage-UI-System-Test/2173/)

https://cloudifysource.atlassian.net/browse/RD-2792

I changed the GenericField component for stateless component and simplified. 

The PR contains also a fix for: https://cloudifysource.atlassian.net/browse/RD-4371

## Checklist

## Tests

 

Stage system tests has been passed with those changes https://jenkins.cloudify.co/job/Stage-UI-System-Test/2173/ [passed]


src/components/form/GenericField/GenericField.tsx
```diff
@@ -214,96 +225,75 @@ class GenericField extends React.PureComponent<GenericFieldProps, GenericFieldSt
...
+            return [
+                { text: 'false', value: false },
+                { text: 'true', value: true }
+            ];
         }
-    }
 
-    handleInputChange: onChange = (proxy, field) => {
-        const { onChange = GenericField.defaultProps.onChange, type } = this.props;
-        onChange(proxy, { ...field, genericType: type });
```
jawamartus, line 237: I removed object field genericType inside of callback method here and in the place where it was in use. https://github.com/cloudify-cosmo/cloudify-stage/pull/1846

It helps me to get rid of handleInputChange method.

I replace it with more clean solution.

src/components/form/GenericField/GenericField.tsx
```diff
@@ -214,96 +225,75 @@ class GenericField extends React.PureComponent<GenericFieldProps, GenericFieldSt
...
-    };
-
-    initOptions(props: GenericFieldProps): void {
-        if (props.type === GenericFieldEnum.BOOLEAN_LIST_TYPE) {
-            this.setState({
-                options: [
```
jawamartus, line 243: We use options only for one or a few types of components. I moved the logic from here to the more appropriate place inside of a new method renderGenericField() .

src/components/form/GenericField/GenericField.tsx
```diff
@@ -120,24 +135,20 @@ interface GenericFieldProps {
...
-    options?: GenericFieldItemType[];
+    // The following statement is neccesery because we use spread operator to pass the rest of props.
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    [x: string]: any;
 }
 
 /**
  * `GenericField` is a generic component which can be used as different input fields in Form component
  */
-class GenericField extends React.PureComponent<GenericFieldProps, GenericFieldState> {
```
jawamartus, line 133: Because of many changes it's hard to read in github it. I recommend to open the whole file in the separated browser tab to see the final result more clearly.

src/components/form/GenericField/GenericField.tsx
```diff
@@ -320,28 +310,37 @@ class GenericField extends React.PureComponent<GenericFieldProps, GenericFieldSt
...
+        }
+        if (type === GenericFieldEnum.CUSTOM_TYPE) {
+            const { default: defaultValue, widgetlessToolbox } = this.props;
             const optionalProps = _.keys(GenericField.defaultProps);
             const requiredProps = ['name', 'label', 'component'];
-            const componentProps = _.omit(this.props, [...optionalProps, ...requiredProps]);
+            const componentProps = _.omit(this.props, ['widgetlessToolbox', ...optionalProps, ...requiredProps]);
 
             if (Component === null || Component === undefined) {
                 return <ErrorMessage error={`\`component\` prop have to be provided when \`${type}\` type is set.`} />;
```
jawamartus, line 325: Previously the component was returning this component in the case of error. Currently it returns <ErrorMEssage /> wrapped with <FormField />. It's more unified solution it doesn't affect end users (only developers during testing). Because of that change I changed one loki test.

src/components/form/GenericField/GenericField.tsx
```diff
@@ -16,15 +16,28 @@ type GenericFieldItemType = {
...
 
 export interface ComponentProps {
     name: string;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
     value: any;
     onChange: onChange;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    widgetlessToolbox: any;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    [x: string]: any;
```
kubama, line 37: Wouldn't it be better to extend `Record<string, any>`?

src/components/form/GenericField/GenericField.tsx
```diff
@@ -16,15 +16,28 @@ type GenericFieldItemType = {
...
+    }
+) => void;
 
 export interface ComponentProps {
     name: string;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
     value: any;
     onChange: onChange;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    widgetlessToolbox: any;
```
kubama, line 35: The concept of widget or widget toolbox is stage specific. If so, then the whole  `GenericField` component is now stage specific. I think it should be moved to stage then. One immediate benefit would be that we could use type for `widgetlessToolbox` that is specified there.

src/components/form/GenericField/GenericField.tsx
kubama: I'm not sure I understand this comment.
Looking at loki screenshot update I can see that 'No component prop specified' is now rendered as part of the component. I think a suitable place for such messages is log/browser console, not the actual application screen.

src/components/form/GenericField/GenericField.tsx
jawamartus: I discussed that with Jakub N. @qooban two days ago. I decided to keep it here, because it would take another week for me to move it to stage. What is more, after moving to stage we will lose loki tests and storybook description inside of stage.  It's not worth the effort right now.

jawamartus: However I like that you noticed that.

kubama: Ok, lets leave it then. Are we going to have a tech-debt ticket for that?

src/components/form/GenericField/GenericField.tsx
qooban: Good idea 👍🏻

src/components/form/Checkbox/Checkbox.tsx
```diff
@@ -8,7 +8,7 @@ export interface CheckboxProps extends FormCheckboxProps {
     /**
      * checkbox label
      */
-    label: string;
+    label: ReactNode;
```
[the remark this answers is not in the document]
jawamartus: Exactly! It is used as ReactNode. Sometimes it can be a string and sometimes a React component.

src/components/form/GenericField/GenericField.stories.tsx
```diff
@@ -16,7 +19,7 @@ export default {
     decorators: [LiveEditDecorator({ DivContainer, GenericField, Form })]
 };
 
-export const stringField = () => {
+export const stringField: Story<Required<GenericFieldProps>> = () => {
```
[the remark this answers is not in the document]
jawamartus: I think it's OK without additional variable. We are using it this way in other storybook stories files.

test/__snapshots__/HtmlSnapshots.test.js.snap
```diff
@@ -11861,22 +11861,30 @@ exports[`Storyshots Form/GenericField Custom Field 1`] = `
...
-  className="ui error message"
+  className="field timeFilterTest"
 >
-  <i
-    aria-hidden="true"
-    className="close icon"
-    onClick={[Function]}
-  />
+  <label>
+    No component prop specified
```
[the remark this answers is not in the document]
jawamartus: AFAIK there is no expected result for this case. It's the exception visible only for `cloudify-stage` developers I hope..

src/components/form/GenericField/GenericField.tsx
jawamartus: Is there any scenario when the <ErrorMessage /> component is visible for the end-users?

src/components/form/GenericField/GenericField.tsx
jawamartus: The result is the same. We don't need to extend Record<string, any>. I prefer this more flat way without inheritance. That's how SemanticReactUI library is doing it.

src/components/form/GenericField/GenericField.tsx
jawamartus: https://cloudifysource.atlassian.net/browse/RD-4419

src/components/form/GenericField/GenericField.tsx
kubama: I think `Record<string, any>` would make lint rule disable unnecessary, that's why I think it would be better here.

jawamartus: For me trying to fit eslint rules it's not the reason to introduce redundant inheritance. It's not for what inheritance is.

kubama: I don't think I get it, where do you see redundancy in inheritance approach?

src/components/form/GenericField/GenericField.stories.tsx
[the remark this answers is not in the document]
qooban: I think it's against DRY and "using it this way in other storybook stories files" argument does not convince me, but as it's Optional, I'm OK with leaving it as is.

src/components/form/GenericField/GenericField.tsx
qooban: > So I think a proper direction should be opposite than the one taken - I think the ErrorMessage display on missing component prop should be removed along with the additional label added in this PR. Instead a more strict props interface should be used.

Sounds like a good idea 👍🏻

qooban (commented):
Looks quite good, I don't have more comments than already opened. 
@jawamartus Please work with @kubama to resolve leftovers and then we can merge it.

src/components/form/GenericField/GenericField.tsx
jawamartus: Good catch! I like the idea of making typing more strict in the future and getting rid of this strange behaviour existing right now. I think that it can be a subtask or another point at: https://cloudifysource.atlassian.net/browse/RD-4419 Would you mind to describe it there?

Definitely it should be a part of another PR. I thing that described change would take around 2 days of development and testing for me. It's very hard to test cross-repository changes like that.

Is that a blocker somehow for approving PR?

src/components/form/GenericField/GenericField.tsx
jawamartus: We can have a call and discuss the redundant inheritance consequences if you wish. 

Is that part of the code in your opinion somehow improper and blocking this PR?

test/__snapshots__/HtmlSnapshots.test.js.snap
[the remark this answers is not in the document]
qooban: If that's only for the developers, then we should put it in console log possibly.

src/components/form/GenericField/GenericField.tsx
kubama: No, it's not a blocker. But I have to say I'm confused by the discussion. 'Redundant' means excessive, being already covered by something else. I don't understand how this applies to inheritance in this context - could you please elaborate shortly on this?

test/__snapshots__/HtmlSnapshots.test.js.snap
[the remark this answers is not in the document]
jawamartus: Please take a look at my comment here: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/125#discussion_r819361145
.

I think it's a good idea for another task. I think it would take around 2 days for me to change to change the form of error messaging for developers. I propose it as at least separated PR.

src/components/form/GenericField/GenericField.tsx
jawamartus: For sure. 

Did you mean by your comment to write somethiong like?
`export interface ComponentProps extends  Record<string, any>{`

src/components/form/GenericField/GenericField.tsx
kubama: I extended description of RD-4419 to cover this, so I'm resolving this thread.

src/components/form/GenericField/GenericField.tsx
kubama: Yes.

kubama:
@jawamartus Are you planning to address https://github.com/cloudify-cosmo/cloudify-stage/pull/1846#discussion_r818687455 in this PR?

test/__snapshots__/HtmlSnapshots.test.js.snap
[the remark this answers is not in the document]
qooban: OK, let's leave it as is.

src/components/form/GenericField/GenericField.tsx
```diff
@@ -185,19 +196,19 @@ class GenericField extends React.PureComponent<GenericFieldProps, GenericFieldSt
...
 
-    static formatValue = (type: GenericFieldEnum, value: any): boolean | string[] | number => {
+    static formatValue = (type: GenericFieldEnum, value: boolean | string): boolean | string | string[] | number => {
         if (type === GenericFieldEnum.MULTI_SELECT_LIST_TYPE) {
-            return _.split(value, ',');
+            return _.split(String(value), ',');
         }
         if (type === GenericFieldEnum.BOOLEAN_TYPE) {
-            return (_.isBoolean(value) && value) || (_.isString(value) && value === 'true');
+            return (_.isBoolean(value) && value) || (_.isString(value) && String(value) === 'true');
```
kubama, line 204:
```suggestion
            return (_.isBoolean(value) && value) || (typeof value === 'string' && value === 'true');
```
IMO looks just a little bit more reasonable than converting string to string.

src/components/form/GenericField/GenericField.tsx
jawamartus: Tell me then which code is more simple to read and more simple to simpler to process for typescript?

1.
```
Interface AnyProps extends Recrod<string, any> {

}

interface ComponentProps extends AnyProps {
    name: string;
    onChange: onChange;
    // eslint-disable-next-line @typescript-eslint/no-explicit-any
    value: any;
    // eslint-disable-next-line @typescript-eslint/no-explicit-any
    widgetlessToolbox: any;
}

```

OR
2.
```

interface ComponentProps {
    name: string;
    onChange: onChange;
    // eslint-disable-next-line @typescript-eslint/no-explicit-any
    value: any;
    // eslint-disable-next-line @typescript-eslint/no-explicit-any
    widgetlessToolbox: any;

    // eslint-disable-next-line @typescript-eslint/no-explicit-any
    [x: string]: any;
}
```
?

Inheritance is very complex for machines and for humans. I think that it's redundant here. We should limit it and use when it's necessary to make our inheritance chain as short as it makes sense.

Worth to note that I used DRY principle in snippet 1. BUT it doesn't make sense to use it there. Rules are good but sometimes it's good to break them.

I understand that you can have different opinion.

src/components/form/GenericField/GenericField.tsx
jawamartus: Good point! I committed your suggestion.

src/components/form/GenericField/GenericField.tsx
kubama: No idea why should I care about what is simpler to process by TS 🙂. I don't agree that 'inheritance is very complex for machines and for humans' in this context - this case is as trivial as it could be. You also still didn't explain where is the redundancy and I still can't see it. The option 1. that you provided is not exactly what I mean, there is no need for intermediate interface and it could be simplified to:
```
interface ComponentProps extends Record<string, any> {
    name: string;
    onChange: onChange;
    // eslint-disable-next-line @typescript-eslint/no-explicit-any
    value: any;
    // eslint-disable-next-line @typescript-eslint/no-explicit-any
    widgetlessToolbox: any;
}
```
Answering your question - it's equally simple to read but it's shorter and saves on rule disable.

jawamartus:
I changed GenericFieldEnum to GenericFieldType. Is there anything else what blocks you from approving the PR?

kubama (approved):
> I changed GenericFieldEnum to GenericFieldType. Is there anything else what blocks you from approving the PR?

I appreciate addressing this.

There is https://github.com/cloudify-cosmo/cloudify-ui-components/pull/125#discussion_r819444506 still unresolved. I would appreciate not leaving comments unanswered, but since it's a minor thing I'm approving this PR regardless of the final discussion outcome.

src/components/form/GenericField/GenericField.tsx
jawamartus: According to dictionary:
redundant - ([especially](https://dictionary.cambridge.org/dictionary/english/especially) of a word, phrase, etc.) [unnecessary](https://dictionary.cambridge.org/dictionary/english/unnecessary) because it is more than is [needed](https://dictionary.cambridge.org/dictionary/english/needed)

I think that using inheritance here is more than we need. It's too big calibre for trivial problem.

Let imagine that there would be a need to inheritance from some other interface here. It would double complexity because you cannot extends more than one thing(interface, generic, class etc.) in typescipt,

jawamartus:
> There is https://github.com/cloudify-cosmo/cloudify-ui-components/pull/125#discussion_r819444506 still unresolved. I would appreciate not leaving comments unanswered, but since it's a minor thing I'm approving this PR regardless of the final discussion outcome.

I addressed your opinion.

jawamartus:
@qooban I still didn't get your approval. What is a reason for not approving the PR? Is there anything what I can improve to change your mind?

qooban (approved):
> @qooban I still didn't get your approval. What is a reason for not approving the PR? Is there anything what I can improve to change your mind?

The reason is simple - I was offline 🙂

PR #148, opened 2022-08-09
RD-5328: Ensured that the peer dependencies are being supported

Vorbert-Kruk:
## Description

Changes consists of adding `.npmrc` file which enables `legacy-peer-deps` during the execution of `npm install` command
Till this time, we had to implicitly use it while installing new dependencies, to avoid the error coming from `>= 8.6.x npm version`

The alternative would be to update bunch of packages and even bump some inner packages of `cloudify-ui-common` (most of them would need to be bumped by a major version)
That's why I've decided to go with the first approach as it's simplier and more bulletproof 🎉 

## Screenshots / Videos

N/A

## Checklist

## Tests

 

N/A

## Documentation

N/A


Vorbert-Kruk:
Closed as progression on an alternative solution has been started 🚀

PR #165, opened 2022-10-04
RD-5866: Adjusted popup offset

Vorbert-Kruk:
## Description

Changes are consisting of adjusting `<Popup />` offset on the X axis.

I'm planning on releasing new `cloudify-ui-components` version, once the scope of [RD-5715](https://cloudifysource.atlassian.net/browse/RD-5715) would be delivered

### Before
Before

### After
After

## Checklist

## Tests

 

N/A

## Documentation

N/A


qooban (dismissed):
Change looks good. 

Regarding tests:
1. I would check if it works fine in Stage or Composer. Is that popup always extending in top right direction? 🤔 
2. Interesting that there's no Loki reference screenshot to fix.

Vorbert-Kruk:
> I would check if it works fine in Stage or Composer. Is that popup always extending in top right direction? 🤔

I've checked both Stage and Composer repos and it seems like none of them are using a `direction` parameter, which can modify the placement of the popup, so I think that we are good to go with using it 💪

qooban (approved):

PR #171, opened 2022-10-14
RD-5719 Ported table to typescript

Blewin:
## Description

Typescript port of table components

few simplifications, and two places needed linter skips - lets discuss them
## Screenshots / Videos

n/a

## Checklist

## Tests

 

refactor, so existing tests are my guardians 
## Documentation

few better comments here and there but in general, its ok


Vorbert-Kruk:
It looks like the pipeline is not passing 😁 
We can see that [here](https://jenkins.cloudify.co/blue/organizations/jenkins/Component-UI-Multibranch/detail/RD-5719/1/pipeline)

It seems that there are some unused `@ts-expect-error` directives and some incompatible types

src/components/data/DataTable/DataTable.tsx
```diff
@@ -17,6 +18,80 @@ import DataTableContext from './DataTableContext';
...
+export interface DataTableProps {
+    /**
+     * table content
+     */
+    // children: PropTypes.node.isRequired,
+    children: ReactNode[];
+    /**
+     * function used to fetch table data
+     */
+    fetchData: (gridParams: unknown) => Promise<unknown>;
```
Vorbert-Kruk, line 30: This prop was previously marked as **not required**
With this change it will be marked as **required**

It was typed like that consciously or by mistake? 🤔

src/components/data/DataTable/DataTable.tsx
```diff
@@ -17,6 +18,80 @@ import DataTableContext from './DataTableContext';
...
+
+    /**
+     * column name used for data sorting
+     */
+    sortColumn: string;
+
+    /**
+     * true for ascending sort, false for descending sort
+     */
+    sortAscending: boolean;
```
Vorbert-Kruk, line 50: These props weree previously marked as **not required**
With this change they will be marked as **required**

It was typed like that consciously or by mistake? 🤔

src/components/data/DataTable/DataTable.tsx
```diff
@@ -51,11 +126,97 @@ class DataTable extends Component {
...
+        /**
+         * message displayed when there's no data
+         */
+        noDataMessage: PropTypes.node,
+
+        /**
+         * CSS style
+         */
+        style: PropTypes.shape({})
+    };
```
Vorbert-Kruk, line 211: As we are fully migrating component to typescript, we should be able to remove propTypes declaration (as stated in our [UI Code Style](https://cloudifysource.atlassian.net/wiki/spaces/RD/pages/435388492/UI+Code+Style#MUST%3A-Skip-using-propTypes-for-TS-components-used-exclusively-in-fully-migrated-TS-components) 😁)

src/components/data/DataTable/TableRow.tsx
```diff
@@ -17,38 +57,27 @@ import TableDataCell from './TableDataCell';
...
+    const computedChildren: ReactNode[] = [];
 
     const computedStyle = { ...style };
     if (_.isFunction(onClick)) {
         computedStyle.cursor = 'pointer';
     }
 
     React.Children.forEach(children, (child, index) => {
-        if (child && child.type === TableDataCell && showData(index)) {
+        if (child && typeof child === 'object' && 'type' in child && child.type === TableDataCell && showData(index)) {
```
Vorbert-Kruk, line 80: WDYT about the following simplification:

```suggestion
        if (child && child?.type === TableDataCell && showData(index)) {
```

? 🤔

Vorbert-Kruk (commented):
Looks quite neat! 🎉 
Left some comments

src/components/data/DataTable/TableRow.tsx
Blewin: typescript doesnt understand this, sadly, i will try again though

src/components/data/DataTable/TableRow.tsx
```diff
@@ -1,9 +1,49 @@
...
+
+    /**
+     * action to be executed on mouse out event and blur
+     */
+    onMouseOut: TRProps['onMouseOut'] & TRProps['onBlur'];
+
+    /**
+     * CSS style
+     */
+    style: CSSStyleDeclaration;
```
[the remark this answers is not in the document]
Blewin: default props magically turns them not required.

src/components/data/DataTable/DataTable.tsx
Blewin: default props magic!

src/components/data/DataTable/DataTable.tsx
Blewin: default props magic!

src/components/data/DataTable/DataTable.tsx
```diff
@@ -80,29 +239,16 @@ class DataTable extends Component {
...
-            // @ts-expect-error TS(2339) FIXME: Property 'sortAscending' does not exist on type '{... Remove this comment to see the full error message
-            changedProps.sortAscending = sortAscending;
-        }
 
-        if (!_.isEmpty(changedProps)) {
+        if (prevProps.sortColumn !== sortColumn || prevProps.sortAscending !== sortAscending) {
             // eslint-disable-next-line react/no-did-update-set-state
-            this.setState(changedProps);
+            this.setState({ sortColumn, sortAscending });
         }
```
qooban, line 178: It's a bit different than original implementation as we'll always update both state variables no matter if only one or two were modified.

Probably it's fine, but we should double check all is working as expected.

src/components/data/DataTable/TableRow.tsx
```diff
@@ -64,7 +93,6 @@ export default function TableRow({
             onBlur={onMouseOut}
             style={computedStyle}
         >
-            {/* @ts-expect-error TS(7005) FIXME: Variable 'computedChildren' implicitly has an 'any... Remove this comment to see the full error message */}
             {computedChildren}
         </tr>
     );
```
qooban, line 98: I guess we can remove `TableRow.propTypes` and `TableRow.defaultProps` as we are migrating to TS.

PatrycjaKubiczek:
there are already many comments, so I'll wait until it will be resolved and then I will do a review and add mine if required :)

src/components/data/DataTable/DataTable.tsx
Blewin: im pretty sure this is functionally te same, since if a === b then a=b won't do anything, and setstate doesnt keep track of which props it changed.

src/components/data/DataTable/TableRow.tsx
PatrycjaKubiczek: I didn't double check it but maybe try: `if (child && (child as ReactElement).type === TableDataCell && showData(index)) {` maybe that would solve the issue? ofc, we would need to import type ReactElement from React as well

src/components/data/DataTable/DataTable.tsx
qooban: As I said probably it's fine, so OK, let's leave it 🙂

src/components/data/DataTable/DataTable.tsx
```diff
@@ -51,11 +126,97 @@ class DataTable extends Component {
...
+         * CSS style
+         */
+        style: PropTypes.shape({})
+    };
+
+    private paginationRef: React.RefObject<Pagination>;
+
+    private debouncedSearch;
+
+    constructor(props: DataTableProps, context: unknown) {
```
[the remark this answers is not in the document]
qooban: As it's optional, I'm not going to force removing it, let's keep it then.

src/components/data/DataTable/TableRow.tsx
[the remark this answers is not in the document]
Blewin: we talked about this in a meeting and arrived at a consensus

src/components/data/DataTable/DataTable.stories.tsx
```diff
@@ -76,7 +76,6 @@ export const pagination = () => {
 export const selectable = () => {
```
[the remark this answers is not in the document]
Blewin: I thik that already migrated files are not typed correctly, please confirm

src/components/data/DataTable/TableRow.tsx
Blewin: typically I avoid "as", but here it is excellent choice, thank you

qooban:
Need to drop, will review tomorrow.

Just one friendly reminder - in the meantime, please check if all discussions are answered 🙂

src/components/data/DataTable/DataTable.stories.tsx
```diff
@@ -1,9 +1,8 @@
 import React from 'react';
 import _ from 'lodash';
```
PatrycjaKubiczek, line 2: optional: using explicit import instead global variables is one of our current refactors in UI https://cloudifysource.atlassian.net/wiki/spaces/RD/pages/2148073473/Ongoing+Refactors+in+Cloudify+UI#Using-explicit-imports-in-widgets :) 
```suggestion
import { map } from 'lodash';
```

src/components/data/DataTable/DataTable.tsx
```diff
@@ -17,6 +17,108 @@ import DataTableContext from './DataTableContext';
 
 import './DataTable.css';
 
+type FetchDataFunction = (p: {
```
PatrycjaKubiczek, line 20: Could we use something more specific than `p` to increase readability? WDYT?

src/components/data/DataTable/TableDataExpandable.tsx
```diff
@@ -1,18 +1,31 @@
...
+     * number of columns to be spanned
+     */
+    numberOfColumns?: number;
+}
+
 /**
  * Defines content of expandable row in data table. See `TableRowExpandable`.
  *
- * All props are passed to the underlaying `td` element.
+ * All props are passed to the underlying `td` element.
```
PatrycjaKubiczek, line 22: nice catch 🎉

src/components/data/DataTable/DataTable.tsx
Vorbert-Kruk: Nice catch, +1 on that 💪  
`parameters` should be a sufficient name 🎉

Vorbert-Kruk (commented):
Great piece of work 💪 
There are still some not addressed discussions from previous code review
For the future notice, please address all of the discussions before asking for a re-review

But as I've looked on the code, I don't see anything requiring additional comments :)

src/components/data/DataTable/DataTable.stories.tsx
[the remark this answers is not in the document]
qooban: Could you elaborate more on that? If we are typing stories incorrectly, then let's understand how to do it properly. 

Once I understand it I can create a ticket to fix all stories at once in the near future.

qooban (commented):
Answered in one of the ongoing discussions.

test/DataTable.test.tsx
```diff
@@ -13,7 +13,7 @@ describe('<DataTable />', () => {
         <DataTable.Column key="col3" label="Column three" show={false} />,
         [{ k: 1 }, { k: 2 }, { k: 3, s: true }, { k: 4 }, { k: 5 }].map(item => {
             return (
-                <DataTable.Row key={item.k} selected={item.s} onClick={item.s ? clickRowMock : null}>
+                <DataTable.Row key={item.k} selected={item.s} onClick={item.s ? clickRowMock : undefined}>
```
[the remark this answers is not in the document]
Blewin: because that way onClick can be typed only as optional property, and that means if we dont want to define it, we need to use `undefined`

src/components/data/DataTable/DataTable.stories.tsx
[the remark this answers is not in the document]
Blewin: They use proper `Story` type, but they assign it a parameter of arguments for the story, which then they dont use i think just using `Story` is better typing

src/components/data/DataTable/TableRowExpandable.tsx
```diff
@@ -16,17 +17,21 @@ import PropTypes from 'prop-types';
...
      */
+    // eslint-disable-next-line react/no-unused-prop-types
     children: PropTypes.node.isRequired,
 
     /**
      * if true, then expandable part of the row will be shown
      */
+    // eslint-disable-next-line react/no-unused-prop-types
     expanded: PropTypes.bool
 };
```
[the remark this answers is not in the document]
Blewin: its removed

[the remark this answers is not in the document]
Blewin: function component will work, and specifying children will make them mandatory

src/components/data/DataTable/TableFilter.tsx
```diff
@@ -25,23 +22,5 @@ import { Form } from 'semantic-ui-react';
...
-    /**
-     * name of the style class to be added
-     */
-    className: PropTypes.string
-};
 
-TableFilter.defaultProps = {
-    className: ''
-};
+export default Form.Field;
```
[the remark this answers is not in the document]
Blewin: empty fields are allowed :)

src/components/data/DataTable/DataTable.stories.tsx
Blewin: thank you! good one

Blewin:
sorry i've missed some comment fixes, its a long one, im going over all of them again to be sure.

src/components/data/DataTable/TableRowExpandable.tsx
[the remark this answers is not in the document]
Blewin: I've missed this comment before bc  i didnt notice that TableDataExpandable is different from TableRowExpandable :)

src/components/data/DataTable/TableFilter.tsx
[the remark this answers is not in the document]
Blewin: i don't see any reason not to trust the original typings from library, and our use doesnt require children

Blewin:
that should be all!

src/components/data/DataTable/TableFilter.tsx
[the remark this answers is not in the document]
Vorbert-Kruk: This seems to me like a good argument 💪 
I wanted just to confirm, that the typing modification wouldn't brake anything
Thank for providing that confirmation 🎸

src/components/data/DataTable/TableRow.tsx
```diff
@@ -1,9 +1,53 @@
...
 import React from 'react';
-import PropTypes from 'prop-types';
-import _ from 'lodash';
+import { isFunction } from 'lodash';
 
 import TableDataCell from './TableDataCell';
 
+type TRProps = React.HTMLAttributes<HTMLTableRowElement>;
+
+export interface TableRowProps {
```
Vorbert-Kruk, line 9: ### Optional

WDYT about doing 

```typescript
export interface TableRowProps extends React.HTMLAttributes<HTMLTableRowElement> {
```

? 🤔 

With that approach we could refer to the whole typing in external repos 
(`stage`, `composer`), while having a need of referencing it

Right now importing `TableRowProps` seems like a partial solution 😁 
(we'd need to dig into this file and see what kind of type should we refer to)

Vorbert-Kruk (approved):
Looks great!

Left some optional comments, not to block this PR 😉

src/components/data/DataTable/TableRow.tsx
Blewin: i started typing whole long response, why i chose this way, then i noticed... its just bad typing. It doesnt pass the props to TR so it only accepts the props which are written here, eveyrthing else is ignored. :D

src/components/data/DataTable/TableRowExpandable.tsx
```diff
@@ -16,20 +17,23 @@ import PropTypes from 'prop-types';
...
+}
+const TableRowExpandable: FunctionComponent<TableRowExpandableProps> = () => {
+    return <></>;
 };
 
+export default TableRowExpandable;
+
 TableRowExpandable.defaultProps = {
     expanded: false
 };
```
qooban, line 39: Do we need this?

src/components/data/DataTable/TableRowExpandable.tsx
```diff
@@ -16,20 +17,23 @@ import PropTypes from 'prop-types';
...
-    children: PropTypes.node.isRequired,
+    children: ReactNode[];
 
     /**
      * if true, then expandable part of the row will be shown
      */
-    expanded: PropTypes.bool
+    expanded: boolean;
+}
+const TableRowExpandable: FunctionComponent<TableRowExpandableProps> = () => {
```
qooban, line 31: Shouldn't we have
```suggestion
const TableRowExpandable: FunctionComponent<TableRowExpandableProps> = ({ expanded = false }) => {
```
here? 🤔

src/components/data/DataTable/TableDataExpandable.tsx
```diff
@@ -1,18 +1,31 @@
...
+export interface TableDataExpandableProps {
+    /**
+     * expandable row contents
+     */
+    children: ReactNode;
+    /**
+     * number of columns to be spanned
+     */
+    numberOfColumns?: number;
+}
```
qooban, line 17: You missed `className` here.

src/components/data/DataTable/DataTable.tsx
```diff
@@ -17,6 +17,108 @@ import DataTableContext from './DataTableContext';
...
+    totalSize: -1,
+    pageSize: 0,
+    sortColumn: '',
+    sortAscending: true,
+    searchable: false,
+    selectable: false,
+    className: '',
+    noDataAvailable: false,
+    sizeMultiplier: 5,
+    noDataMessage: 'No data available' as ReactNode,
```
qooban, line 41: Out of curiosity - do we need that casting? Isn't `string` part of `ReactNode` type? 🤔

src/components/data/DataTable/DataTable.tsx
```diff
@@ -17,6 +17,108 @@ import DataTableContext from './DataTableContext';
...
+        _search: string;
+        currentPage: number;
+        pageSize: number;
+        sortColumn: string;
+        sortAscending: boolean;
+    };
+}) => void;
+
+const defaultProps = {
+    fetchData: (() => {}) as FetchDataFunction,
```
qooban, line 31: Hmm... In the original prop type definition it returns Promise, so are we OK here? 🤔

Vorbert-Kruk (approved):
LGTM 🎉

src/components/data/DataTable/DataTable.tsx
```diff
@@ -17,6 +17,108 @@ import DataTableContext from './DataTableContext';
...
+     */
+    noDataMessage?: ReactNode;
+
+    /**
+     * CSS style
+     */
+    style?: React.CSSProperties;
+}
+
+type DataTablePropsAfterDefault = DataTableProps & typeof defaultProps;
```
qooban, line 112: Out of curiosity - why do we need that? Shouldn't types for default props and for props be the same?

src/components/data/DataTable/TableDataExpandable.tsx
Blewin: good catch, i missed whole TDProps here tbh

src/components/data/DataTable/TableRowExpandable.tsx
Blewin: in this case, this is not the case, since this render function is never actually run.

In general, i tried to not touch this part very much, since its a bit tricky (this component just acts as a vessel for children, and is unpacked and not rendered when its used properly (inside as table))

src/components/data/DataTable/TableRowExpandable.tsx
Blewin: 99% sure its not needed, but: 

In general, i tried to not touch this part very much, since its a bit tricky (this component just acts as a vessel for children, and is unpacked and not rendered when its used properly (inside as table))

src/components/data/DataTable/DataTable.tsx
Blewin: we don't use it as a promise, and few times in stage we pass not function that doesn't return a promise, so this typing is more true to the reality

src/components/data/DataTable/DataTable.tsx
Blewin: they differ on which props are optional, and which are required

This is type after default - for usage inside the component, so with knowledge that default propsa are provided.

src/components/data/DataTable/DataTable.tsx
Blewin: its so that `type DataTablePropsAfterDefault = DataTableProps & typeof defaultProps;` works properly

src/components/data/DataTable/DataTable.stories.tsx
PatrycjaKubiczek: could I ask where do you getting errors/issues? 🤔 because I tried to run storybook locally after changing lodash imports and replacing `_.map` to `map` and I don't see any issue, but maybe I didn't notice something. It would be useful information for future migration work as well :)

PatrycjaKubiczek: and thanks for adjustments in other places 🎉

src/components/data/DataTable/DataTable.tsx
qooban: OK.

src/components/data/DataTable/DataTable.tsx
qooban: OK.

src/components/data/DataTable/DataTable.tsx
qooban: OK.

qooban (approved):
Looks good 👍🏻

src/components/data/DataTable/TableRowExpandable.tsx
qooban: OK, let's leave it then.

PR #172, opened 2022-10-17
RD-5721: Migrated form components to typescript - Part 2

Vorbert-Kruk:
## Description

Changes consists of:
- fully migrating components (and corresponding tests) to typescript - additionally I've migrated one component test that wasn't migrated within previous PRs
- migrating away from implicit lodash imports, in order of using explicit ones
- removing redundant `@ts-expect-error` comments 
- extracted and shared type between input components 🎉 

## Screenshots / Videos

N/A

## Checklist

## Tests

 

I've checked on the [stage](https://github.com/cloudify-cosmo/cloudify-stage) if everything is working correctly it seems, that 2  typing adjustments will have to be made (due to `Form.fieldNameValue` type modification) - as I'll be responsible for the upcoming `cloudify-ui-components` release, I'll make those adjustments 💪 

As within this PR functional changes weren't made, I didn't go through [stage](https://github.com/cloudify-cosmo/cloudify-stage) application to click over every migrated component 
If it would be considered as a requirement from a reviewer perspective, please let me know :)

## Documentation

N/A


src/components/form/Form/Form.tsx
```diff
@@ -160,7 +160,7 @@ function Form({
     );
 }
 
-Form.fieldNameValue = (field: { name: string; value: unknown; type: string; checked?: string }) => {
+Form.fieldNameValue = (field: { name: string; value: unknown; type: string; checked?: boolean }) => {
```
qooban, line 163: Good catch 👍🏻

qooban, line 163: I'm wondering - I saw in Stage in two places:
```
@ts-expect-error Form.fieldNameValue is not converted to TS yet
```
Will we be able to remove it? 🤔 
(asking as this file was already migrated to TS, so maybe something else is missing)

qooban (commented):
Looks very good 👍🏻 

Added just one command to double-check `fieldNameValue` and I'm wondering if it's about the same issue as you described in `Tests` section:

> I've checked on the [stage](https://github.com/cloudify-cosmo/cloudify-stage) if everything is working correctly it seems, that 2 typing adjustments will have to be made (due to Form.fieldNameValue type modification)

src/components/form/Form/Form.tsx
Vorbert-Kruk: In the description I was referring to other places 😁 
I've checked mentioned by you occurences and it seems that removing those comments would require some additional work

As an investigation notes I can add, that the issues are coming from:
- Using results of `Form.fieldNameValue` to update component state (types seems to incompatible 😁)  
- Incompability of field type passed down to `Form.fieldNameValue` 

I've got an idea how this issue can be approached (in both places) and I'll try to apply it :)
If it will turn out that the fix will be more complicated / would require some bigger amount of work, I'll skip it and leave it for the future adventures 😁

qooban: OK, one small request, reg. "If it will turn out that the fix will be more complicated / would require some bigger amount of work" - in that case please create a ticket and add it to RD-5590 epic.

qooban (approved):

src/components/form/Form/Form.tsx
Vorbert-Kruk: Sure, seems reasonable :)

PR #207, opened 2023-04-05
RND-311 Add `UnmaskableSecretInput` component and `divider` prop

kubama:
## Description
Added stage extracted `UnmaskableSecretInput` (renamed from `SinglelineInput`).
Added `divider` prop to `AccordionSection` component (to be used in place of `AccordionSectionWithDivider` currently in stage).

## Screenshots / Videos
image
image

## Checklist

## Tests
Added snapshots and loki images

 

## Documentation
Added JSDocs


src/components/layout/AccordionSectionWithDivider/AccordionSectionWithDivider.tsx
```diff
@@ -0,0 +1,38 @@
...
+            activeSection={activeSection}
+            onClick={onClick}
+        >
+            <Divider style={dividerStyle} />
+            {children}
+        </AccordionSection>
+    );
+}
+
+export default AccordionSectionWithDivider;
```
Vorbert-Kruk, line 38: ### Suggestion

This component is providing the same functionality as `AccordionSection`, but it's enhancing it by providing `Divider` component

Looking at the name of the component (`AccordionSectionWithDivider`) and type of adjustment that was made, I was wondering if we could just extend `AccordionSection` component with `withDivider` / `showDivider` etc. kind of flag, which would allow to present `Divider` component 🎉 


### Investigation notes
[AccordionSection](https://github.com/cloudify-cosmo/cloudify-ui-components/blob/master/src/components/layout/AccordionSection/AccordionSection.tsx) looks like a custom component (not the one imported from `semantic-ui`), which is an additional benefit from the perspective of implementing suggestion mentioned above 



What do you think about this idea? 🤔

kubama: Actually I was considering it myself, and as you are making the same point I'm now convinced that it's the way to go. I'll implement this change then.

kubama: Done

PatrycjaKubiczek (commented):
unsmask
shouldn't we have some description here? 🤔

PatrycjaKubiczek (approved):
thanks for improving docs, LGTM 👍

Vorbert-Kruk (approved):
LGTM 🎉

PR #213, opened 2023-04-25
RD-6339 Enhance documentation of storybook story definition

PatrycjaKubiczek:
## Description

- remove <story>.storyName = 'Default'
- update snapshots - it seems like a lot of changed - but that is because of changes in export naming for example earlier we had:
 ```
exports[`Storyshots Buttons/ApproveButton Default 1`] 
```
and now it's:
```
exports[`Storyshots Buttons/ApproveButton Basic 1`] 
```
 and so it changed the order of components in a snapshot, but the components themselves didn't change
- same with the visual screenshots - it's different naming now, and there are visible changes in some - but it seems like the screenshot didn't update earlier even it was supposed to update earlier (which is a bit weird  I know) - so we're all good - I know it looks confusing, it looked confusing for me as well in the beginning but I checked components and they look like supposed to look
- created a template that could be used as a base for the creation of the future *.stories.tsx files - however initially I wanted to put it `components` but this creates an additional component visible in the Storybook - so I moved that to `src` I'm open to hearing any suggestion because I didn't have any better idea how to do it :) 

## Screenshots / Videos

n/a

## Checklist

## Tests

 

CI checks are green ✔️ 

## Documentation

n/a


kubama:
> same with the visual screenshots - it's different naming now, and there are visible changes in some - but it seems like the screenshot didn't update earlier even it was supposed to update earlier (which is a bit weird I know) - so we're all good - I know it looks confusing, it looked confusing for me as well in the beginning but I checked components and they look like supposed to look

Nothing confusing about it - loki images are compared with a certain level of tolerance. So it's not an unexpected scenario that a component was changed visually in the past, its image was not updated (intentionally or not) and the build continued to pass.

> created a template that could be used as a base for the creation of the future *.stories.tsx files - however initially I wanted to put it components but this creates an additional component visible in the Storybook - so I moved that to src I'm open to hearing any suggestion because I didn't have any better idea how to do it :)

I'm not sure how useful it is. I think it's extremely easy to start off by copying from an existing simple component, like one of the buttons. Personally I'd rather drop it - as it's not production code it can easily fall out of sync with the rest of the code base.

PatrycjaKubiczek:
@Vorbert-Kruk as you are the reporter of this ticket - what do you think about dropping the template? As we've already talked earlier we had some hesitation about this feature anyway :)

Vorbert-Kruk:
> @Vorbert-Kruk as you are the reporter of this ticket - what do you think about dropping the template? As we've already talked earlier we had some hesitation about this feature anyway :)

Sure, I'm totally fine with it 😉

kubama (approved):

PR #219, opened 2023-05-11
RD-459 Create Text Input component

PatrycjaKubiczek:
## Description

- created Text Input component 

notes: 
- icon is slightly lower than it should it because we have a different rem size than DDS - I wasn't sure if I should fix that or not 🤔
- I added a form tag to the text input to mock input with error (it's necessary to have proper styling), IMO Form should be a separate component  

## Screenshots / Videos

Screenshot 2023-05-11 162658

## Checklist

## Tests

 

CI checks ✔️ 

## Documentation

n/a


kubama:
> * I added a form tag to the text input to mock input with error (it's necessary to have proper styling), IMO Form should be a separate component

How about leaving TODO to address it once `Form` component is available?

src/components/dds/textInput/TextInput.tsx
```diff
@@ -0,0 +1,153 @@
...
+    helperText,
+    value,
+    onChange,
+    ...rest
+}: TextInputProps) {
+    const isPasswordType = type === 'password';
+    const [isPasswordVisible, toggleInputPassword] = useToggle(false);
+
+    return (
+        <div className={classNames(inputClasses.container, errorText ? inputClasses.isInvalid : '')} {...rest}>
```
kubama, line 109: What will happen if `className` is specified in `rest`?

kubama: Ok, I can see it will never be the case :).

kubama, line 109: With `classNames` we can save few keystrokes as it handles `false` value:
```suggestion
        <div className={classNames(inputClasses.container, errorText && inputClasses.isInvalid)} {...rest}>
```

src/components/dds/textInput/TextInput.tsx
```diff
@@ -0,0 +1,153 @@
...
+                    required={required}
+                    disabled={inactive}
+                    value={value}
+                    onChange={onChange}
+                />
+                {iconName && (
+                    <span
+                        className={classNames(inputClasses.icon, inputClasses.iconStart, `dds__icon--${iconName}`)}
+                        aria-hidden="true"
+                    />
```
kubama, line 140: Can't we use `Icon` component here?

src/components/dds/textInput/TextInput.tsx
```diff
@@ -0,0 +1,153 @@
...
+                )}
+                {errorText && (
+                    <Icon className={classNames(inputClasses.iconEnd, inputClasses.errorIcon)} name="alert-notice" />
+                )}
+
+                {helperText && <small className={classNames(inputClasses.helperText)}>{helperText}</small>}
+                {errorText && <div className={classNames(inputClasses.errorText)}>{errorText}</div>}
```
kubama, line 147: I think we don't need the wrapping `classNames` invocation.

src/components/dds/textInput/TextInput.tsx
```diff
@@ -0,0 +1,153 @@
...
+                    type="button"
+                    className={classNames(inputClasses.type.password.action, passwordClasses.switchButton)}
+                    aria-label={isPasswordVisible ? defaultProps.srHideTitle : defaultProps.srHideTitle}
+                    onClick={toggleInputPassword}
+                >
+                    {isPasswordVisible ? defaultProps.hideLabel : defaultProps.showLabel}
+                </button>
+            )}
+
+            <div className={classNames(inputClasses.wrapper)}>
```
kubama, line 124: I think we don't need the wrapping `classNames` invocation.

src/components/dds/textInput/TextInput.tsx
```diff
@@ -0,0 +1,153 @@
...
+} from 'dds-core/components/input-password/ts/input-password';
+import 'dds-core/components/text-input/_text-input.scss';
+import 'dds-core/components/label/_label.scss';
+import 'dds-core/components/input-password/_input-password.scss';
+import 'dds-core/components/form/_form.scss';
+import type { Icons, IconType } from 'dds-core/components/icon/ts/icon';
+import { useToggle } from 'hooks';
+import { Icon } from '../icon';
+
+export interface TextInputProps extends React.HTMLAttributes<HTMLDivElement> {
```
kubama, line 18: I'm wondering - maybe it would actually make more sense to extend from props of `input` element? We wouldn't have to specify props that are directly passed there (like `minLength`, `maxLength` and so on). We could just use `{...rest}` on `input` element. It's the `input` element that's at the core of this component after all.

kubama:
Added @qooban to the review as he is also working on DDS

src/components/dds/textInput/TextInput.tsx
PatrycjaKubiczek: fixed: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/219/commits/2426655a61312da027facd7e03b74d48a88213fa

src/components/dds/textInput/TextInput.tsx
PatrycjaKubiczek: fixed: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/219/commits/2426655a61312da027facd7e03b74d48a88213fa

src/components/dds/textInput/TextInput.tsx
PatrycjaKubiczek: changed: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/219/commits/2426655a61312da027facd7e03b74d48a88213fa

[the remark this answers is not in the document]
PatrycjaKubiczek: done: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/219/commits/2426655a61312da027facd7e03b74d48a88213fa

src/components/dds/textInput/TextInput.stories.tsx
```diff
@@ -0,0 +1,45 @@
...
+    <>
+        <TextInput labelText="Username" iconName="user" />
+    </>
+);
+
+export const password: TextInputStory = () => (
+    <>
+        <TextInput labelText="Password" type="password" />
+    </>
+);
```
qooban, line 45: How about removing wrapping `<></>` tags? 🤔 
It would be a bit shorter and cleaner IMHO.

src/components/dds/textInput/TextInput.tsx
```diff
@@ -0,0 +1,145 @@
...
+import 'dds-core/components/input-password/_input-password.scss';
+import 'dds-core/components/form/_form.scss';
+import type { Icons, IconType } from 'dds-core/components/icon/ts/icon';
+import { useToggle } from 'hooks';
+import { Icon } from '../icon';
+
+export interface TextInputProps extends React.InputHTMLAttributes<HTMLInputElement> {
+    /**
+     * 	The input type to use with this component (text|url|search|tel|email)
+     *
```
qooban, line 18: Minor thing, just to have it elegant and clean - could you take a look at props' comments endings (sometimes there are empty lines, sometimes `* */`) and make it consistent?

src/components/dds/textInput/TextInput.tsx
```diff
@@ -0,0 +1,145 @@
...
+     * Define the type of icon that will be added (font-icon|svg)
+     */
+    iconType?: IconType;
+    /**
+     * Whether the text input should be disabled
+     */
+    inactive?: boolean;
+    /**
+     * Toggles the 'invalid' attribute. Used when the component is in a form with validation.
+     * @default false
```
qooban, line 35: I think Storybook in props documentation table automatically takes it from L89 or am I missing something? 🤔

src/components/dds/textInput/TextInput.tsx
```diff
@@ -0,0 +1,145 @@
...
+     * */
+    errorText?: string;
+    /**
+     * Helper text for the text input
+     * */
+    helperText?: string;
+    /**
+     * Specifies the initial value for this component
+     * */
+    value?: any;
```
qooban, line 74: I'm wondering - if it is text input can it be anything different than string? 🤔

src/components/dds/textInput/TextInput.tsx
```diff
@@ -0,0 +1,145 @@
...
+    value,
+    onChange,
+    ...rest
+}: TextInputProps) {
+    const isPasswordType = type === 'password';
+    const [isPasswordVisible, toggleInputPassword] = useToggle(false);
+
+    return (
+        <div className={classNames(inputClasses.container, errorText && inputClasses.isInvalid, className)}>
+            <label htmlFor="input" className={(labelClasses.base, required ? labelClasses.required : '')}>
```
qooban, line 104: Maybe as you used `classNames` in the previous line, here it would work as well:
```suggestion
            <label htmlFor="input" className={classNames(labelClasses.base, required && labelClasses.required)}>
```
?

qooban, line 104: Maybe as you used `classNames` in the previous line, here it would work as well:
```suggestion
            <label htmlFor="input" className={classNames(labelClasses.base, required && labelClasses.required)}>
```
?

src/components/dds/textInput/TextInput.tsx
PatrycjaKubiczek: totally agree, good idea, changed it: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/219/commits/2426655a61312da027facd7e03b74d48a88213fa

qooban:
> icon is slightly lower than it should it because we have a different rem size than DDS - I wasn't sure if I should fix that or not thinking

Let's leave it for now. I experienced the same issue when developing `Card` component.

src/components/dds/textInput/TextInput.stories.tsx
PatrycjaKubiczek: removed: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/219/commits/76ac6a02c0667a6adb27ce7b8eec0f380c9d3639

src/components/dds/textInput/TextInput.tsx
PatrycjaKubiczek: fixed it: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/219/commits/4357781ebb75ad4303a1a11c8f12d2abb38be1bb

src/components/dds/textInput/TextInput.tsx
PatrycjaKubiczek: correct, fixed: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/219/commits/4357781ebb75ad4303a1a11c8f12d2abb38be1bb :)

src/components/dds/textInput/TextInput.tsx
PatrycjaKubiczek: changed it: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/219/commits/07e87ed5af0c09081407561a0e0fbe38fe9a2ef9

PatrycjaKubiczek: I think this comment is duplicate of previous one :) https://github.com/cloudify-cosmo/cloudify-ui-components/pull/219/commits/07e87ed5af0c09081407561a0e0fbe38fe9a2ef9 resolved here

src/components/dds/textInput/TextInput.tsx
```diff
@@ -0,0 +1,142 @@
...
+    value,
+    onChange,
+    ...rest
+}: TextInputProps) {
+    const isPasswordType = type === 'password';
+    const [isPasswordVisible, toggleInputPassword] = useToggle(false);
+
+    return (
+        <div className={classNames(inputClasses.container, errorText && inputClasses.isInvalid, className)}>
+            <label htmlFor="input" className={classNames(labelClasses.base, required && labelClasses.required)}>
```
Vorbert-Kruk, line 101: Looking at the [documentation of htmlFor attribute](https://developer.mozilla.org/en-US/docs/Web/API/HTMLLabelElement/htmlFor) I've stumbled across the following description:

> The HTMLLabelElement.htmlFor property reflects the value of the [for](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#for) content property. That means that this script-accessible property is used to set and read the value of the content property for, which is the ID of the label's associated control element.

So it looks like `htmlFor` attribute is accepting `id` of the element to which we want it to relate
As currently it's being presented with the "`input`" string, I don't think it is being used properly - we are not using "`input`" as an id for anything

### Conclusion

So the question is if it's working properly right now
We could check it by clicking a label and expecting that the related input would be focused after that click
As I'm checking it locally, it doesn't seem to work

So now the question appears if we need that functionality or not
If we don't, we could just remove `htmlFor` attribute
If we would, we could generate unique id, within the component, which we could pass to both:
- `htmlFor` attribute of `label`
- `id` attribute of `input`

src/components/dds/textInput/TextInput.tsx
Vorbert-Kruk: I was also thinking about this suggestion

PatrycjaKubiczek: I thought about it too, but also I wanted to follow how it was prepared in [DDS docs](https://vanilla.delldesignsystem.com/2.28.0/index.html?path=/docs/components-text-input--basic) where we could see it was span.  
Another thing if it's Icon we have a bit of different styling -notice the red color of icon in screenshot: 
Screenshot 2023-05-12 153133

and in DDS in mockups it's black color:
Screenshot 2023-05-12 153305

so WDYT? should I change it to Icon and add black color or leave as it is?

kubama: > I thought about it too, but also I wanted to follow how it was prepared in [DDS docs](https://vanilla.delldesignsystem.com/2.28.0/index.html?path=/docs/components-text-input--basic) where we could see it was span.

I don't think it actually makes any difference. When working on `Icon` I noticed in different places of DDS documentation it was using `span` or `i`.
Another thing is that the `Icon` component makes use of DDS icons font, while your current implementation does not.

> Another thing if it's Icon we have a bit of different styling -notice the red color of icon in screenshot

Why can't we fix it?

src/components/dds/textInput/TextInput.tsx
PatrycjaKubiczek: changed it: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/219/commits/3bb81be2ee0c5a2723ee419cf3c5a926dd4a3a2c

src/components/dds/textInput/TextInput.tsx
PatrycjaKubiczek: okay, I wasn't sure what approach we have to DDS in general - thanks for clearing it out 
https://github.com/cloudify-cosmo/cloudify-ui-components/pull/219/commits/98a407b457df27b1027e9370fe425b25b268842e

do you also know what I should do with colors? are we planning to have them in ui-common repo as well? 🤔

kubama: We have https://cloudifysource.atlassian.net/browse/RND-564 planned.
For now you can hardcode the value if needed and reference RND-564 in the TODO.

src/components/dds/textInput/TextInput.tsx
PatrycjaKubiczek: we have this eslint rule in UI-components - https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/label-has-associated-control.md so I guess removing it wouldn't be the best 🤔

src/components/dds/textInput/TextInput.tsx
PatrycjaKubiczek: done: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/219/commits/b643c179a36c6f916a39efb85821af1e1dc4dd53

src/components/dds/textInput/TextInput.tsx
```diff
@@ -0,0 +1,145 @@
...
+    helperText,
+    value,
+    onChange,
+    ...rest
+}: TextInputProps) {
+    const isPasswordType = type === 'password';
+    const [isPasswordVisible, toggleInputPassword] = useToggle(false);
+
+    return (
+        <div className={classNames(inputClasses.container, errorText && inputClasses.isInvalid, className)}>
```
kubama, line 101: For root level element we should use `getRootClassNames` to create class string. This makes sure DDS reboot wrapper is in place.

src/components/dds/textInput/TextInput.tsx
Vorbert-Kruk: In that case we could go for a first suggestion, which was mentioning possibility of generating unique id for input instances, so that `htmlFor` attribute would be working
We could implement it in the following way:

```typescript
import { uniqueId } from 'lodash';
...
export function TextInput({...}: TextInputProps) {
    const inputId = uniqueId();
    const isPasswordType = type === 'password';
    const [isPasswordVisible, toggleInputPassword] = useToggle(false);

    return (
        <div className={classNames(inputClasses.container, errorText && inputClasses.isInvalid, className)}>
            <label htmlFor={inputId} className={classNames(labelClasses.base, required && labelClasses.required)}>
                {labelText}
            </label>
            {isPasswordType && (
                <button
                    type="button"
                    className={classNames(inputClasses.type.password.action, passwordClasses.switchButton)}
                    aria-label={isPasswordVisible ? defaultProps.srHideTitle : defaultProps.srHideTitle}
                    onClick={toggleInputPassword}
                >
                    {isPasswordVisible ? defaultProps.hideLabel : defaultProps.showLabel}
                </button>
            )}

            <div className={inputClasses.wrapper}>
                <input
                    id={inputId}
                    type={type === 'password' && isPasswordVisible ? 'text' : type}
                    className={classNames(inputClasses.type.text.control, iconName && inputClasses.hasIcon)}
                    required={required}
                    disabled={inactive}
                    value={value}
                    onChange={onChange}
                    {...rest}
                />
                {iconName && (
                    <Icon
                        className={classNames(inputClasses.icon, inputClasses.iconStart)}
                        name={iconName}
                        // TODO RND-564 - change hardcoded color to variable
                        style={{ color: '#636363' }}
                    />
                )}
                {errorText && (
                    <Icon className={classNames(inputClasses.iconEnd, inputClasses.errorIcon)} name="alert-notice" />
                )}

                {helperText && <small className={inputClasses.helperText}>{helperText}</small>}
                {errorText && <div className={inputClasses.errorText}>{errorText}</div>}
            </div>
        </div>
    );
}
```

Let me know WDYT

qooban (dismissed):

src/components/dds/textInput/TextInput.tsx
PatrycjaKubiczek: done: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/219/commits/ae62701984bfd2c0442924208be2595741060ef6

src/components/dds/textInput/TextInput.tsx
PatrycjaKubiczek: looks great, implemented it: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/219/commits/b787e50caf2138f04034e2deae5be52816ca0a43 💪

src/components/dds/textInput/TextInput.tsx
```diff
@@ -0,0 +1,148 @@
...
+     * */
+    helperText?: string;
+    /**
+     * Specifies the initial value for this component
+     * */
+    value?: string;
+    /**
+     * Function called on input value change
+     */
+    onChange?: (event: React.ChangeEvent<HTMLInputElement>) => void;
```
kubama, line 77: For ease of use of this component, how about:
```suggestion
    onChange?: (value: string) => void;
```
or
```suggestion
    onChange?: (value: string, event: React.ChangeEvent<HTMLInputElement>) => void;
```
?
I think every usage of this callback will be extracting the value from the event, so it would be beneficial to receive the value directly.

Alternatively we could stick to signature specific to Semantic UI, as we already have `useInput` and `useInputs` hooks that work with that:
```suggestion
    onChange?: (event: React.ChangeEvent<HTMLInputElement>, data: InputOnChangeData) => void;
```
(however I'm not a fan of such arguments order, I'd prefer to have `data` first, as it is frequently used, and `event` second, as it is frequently ignored)

src/components/dds/textInput/TextInput.tsx
```diff
@@ -0,0 +1,148 @@
...
+import { componentClasses as passwordClasses, defaultProps } from 'dds-core/components/input-password/input-password';
+import 'dds-core/components/text-input/_text-input.scss';
+import 'dds-core/components/label/_label.scss';
+import 'dds-core/components/input-password/_input-password.scss';
+import 'dds-core/components/form/_form.scss';
+import type { Icons, IconType } from 'dds-core/components/icon/ts/icon';
+import { useToggle } from 'hooks';
+import { Icon } from '../icon';
+
+export interface TextInputProps extends React.InputHTMLAttributes<HTMLInputElement> {
```
kubama, line 17: This interface specifies a number of props that are already defined in `React.InputHTMLAttributes`. Does it make sense to repeat them here?

src/components/dds/textInput/TextInput.tsx
```diff
@@ -0,0 +1,148 @@
...
+    value,
+    onChange,
+    ...rest
+}: TextInputProps) {
+    const inputID = uniqueId();
+    const isPasswordType = type === 'password';
+    const [isPasswordVisible, toggleInputPassword] = useToggle(false);
+
+    return (
+        <div className={getRootClassNames(inputClasses.container, errorText && inputClasses.isInvalid, className)}>
```
kubama, line 103: I think it would make sense to set `style` prop on this top-level element instead of `input`. It would enable using `style` for layout related styling.

Vorbert-Kruk (dismissed):
LGTM 🎉 
Thanks for addressing all comments!

src/components/dds/textInput/TextInput.tsx
```diff
@@ -0,0 +1,148 @@
...
+     */
+    type?: Type;
+    /**
+     * 	The name of a DDS icon (without the prefix *dds__*)
+     */
+    iconName?: Icons;
+    /**
+     * Define the type of icon that will be added (font-icon|svg)
+     */
+    iconType?: IconType;
```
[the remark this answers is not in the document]
PatrycjaKubiczek: done: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/219/commits/85421b5b29f4d205e79589a06699bda428481919#diff-fb2ed380c7c73e92270cc6ea2ca196245d7738ab7c1aeb31b7c17b05c5fb7a12

src/components/dds/textInput/TextInput.tsx
PatrycjaKubiczek: done: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/219/commits/e2d1045b44dce89e23b4eb2e758475ec89d4ec48

src/components/dds/textInput/TextInput.tsx
PatrycjaKubiczek: I thought it should be similar to DDS docs so I followed how it looked there, but I removed them: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/219/commits/ee5552868f363ec99a00d47bd5c55c798be4bcdf

src/components/dds/textInput/TextInput.tsx
PatrycjaKubiczek: changed it to second proposition: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/219/commits/e2d1045b44dce89e23b4eb2e758475ec89d4ec48

src/components/dds/textInput/TextInput.tsx
```diff
@@ -10,23 +10,18 @@ import 'dds-core/components/text-input/_text-input.scss';
...
-import type { Icons, IconType } from 'dds-core/components/icon/ts/icon';
+import type { IconDefinitionProps } from '../common/icons';
 import { useToggle } from 'hooks';
 import { Icon } from '../icon';
 
-export interface TextInputProps extends React.InputHTMLAttributes<HTMLInputElement> {
+export interface TextInputProps
+    extends Omit<React.InputHTMLAttributes<HTMLInputElement>, 'onChange'>,
+        IconDefinitionProps {
+    className?: string;
```
kubama: Why do we now need `className` here?

src/components/dds/TextInput/TextInput.tsx
```diff
@@ -123,8 +106,8 @@ export function TextInput({
                     required={required}
                     disabled={inactive}
                     value={value}
-                    onChange={onChange}
-                    {...rest}
+                    onChange={e => onChange?.(e.target.value, e)}
+                    {...omit(rest, ['style'])}
```
kubama, line 110: At this point `rest` doesn't contain `style` due to L75, so we don't need `omit` here.

src/components/dds/TextInput/TextInput.tsx
```diff
@@ -93,14 +72,18 @@ export function TextInput({
...
     ...rest
 }: TextInputProps) {
     const inputID = uniqueId();
     const isPasswordType = type === 'password';
     const [isPasswordVisible, toggleInputPassword] = useToggle(false);
 
     return (
-        <div className={getRootClassNames(inputClasses.container, errorText && inputClasses.isInvalid, className)}>
+        <div
+            className={getRootClassNames(inputClasses.container, errorText && inputClasses.isInvalid, className)}
```
kubama, line 84: The first argument to `getRootClassNames` is `classNameFromProps`, so we could move `className` to the beginning of the arguments list to be in line with that.

src/components/dds/TextInput/TextInput.tsx
```diff
@@ -40,25 +35,9 @@ export interface TextInputProps extends React.InputHTMLAttributes<HTMLInputEleme
...
-    name?: string;
-    /**
-     * 	Provides a brief hint to the user as to what kind of information is expected in the field
-     * */
-    placeholder?: string;
-    /**
-     * 	Toggles the 'required' attribute. Used when the component is in a form with validation.
-     */
-    required?: boolean;
+    style?: React.CSSProperties;
```
kubama, line 40: I think `style` is already part of `React.InputHTMLAttributes`

src/components/dds/textInput/TextInput.tsx
PatrycjaKubiczek: we don't : ) removed: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/219/commits/f2049ed806ccc3e2f28ace21686e439033f4ab27

src/components/dds/TextInput/TextInput.tsx
PatrycjaKubiczek: removed: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/219/commits/f2049ed806ccc3e2f28ace21686e439033f4ab27

src/components/dds/TextInput/TextInput.tsx
PatrycjaKubiczek: changed: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/219/commits/f2049ed806ccc3e2f28ace21686e439033f4ab27

src/components/dds/TextInput/TextInput.tsx
PatrycjaKubiczek: removed 'style': https://github.com/cloudify-cosmo/cloudify-ui-components/pull/219/commits/f2049ed806ccc3e2f28ace21686e439033f4ab27

kubama (approved):

PR #221, opened 2023-05-11
RND-562 Add basic support for DDS Grid foundation

qooban:
## Description

Added basic support for DDS based `Grid` component.

Based on [DDS - Grid foundation](https://www.delldesignsystem.com/foundations/grid/) and [DDS Vanilla - Grid foundation](https://vanilla.delldesignsystem.com/2.28.0/index.html?path=/docs/foundations-grid--page).

Not included for now:
1. Flex and auto columns
2. Offsets
3. Order

described in [DDS Vanilla - Grid foundation](https://vanilla.delldesignsystem.com/2.28.0/index.html?path=/docs/foundations-grid--page).

### Basic screenshot

image

### Video from Storybook

https://github.com/cloudify-cosmo/cloudify-ui-components/assets/5202105/6d83e7c7-c31a-403d-ab28-bab17e80e025

## Checklist
- [x] My code follows the [UI Code Style](https://cloudifysource.atlassian.net/l/c/x8fq902p).
- [x] I followed the [UI Testing Guidelines](https://cloudifysource.atlassian.net/l/cp/udCVfvrx) and verified that all tests/checks have passed.
- [x] I verified if that change requires any update in the [documentation](https://cloudifysource.atlassian.net/l/c/4XKtPocy).
- [x] I added proper labels to this PR.

## Tests
Updated HTML snapshots and added loki reference images.

## Documentation
Added stories and JSDoc.

PR #222, opened 2023-05-12
RND-465 Add DDS Card component

qooban:
## Description

Created DDS `Card` component.
Ref.: https://www.delldesignsystem.com/components/card/

Not included:
1. adding links to footer (DDS Link component is not available yet)
2. optional container for charts or graphs in body (no need to support that basing on the mockups we have)

This PR bases on #221  .

### Basic example

image

### Video from Storybook

https://github.com/cloudify-cosmo/cloudify-ui-components/assets/5202105/f0e9aa40-d9e9-4e21-92ff-69e5b134d6fe

## Checklist

## Tests
Added Loki reference images and HTML snapshots.

## Documentation
Added JSDoc for component props and Storybook stories.


src/components/dds/card/CardFooter.tsx
```diff
@@ -0,0 +1,36 @@
...
+ * Card component footer
+ */
+export function CardFooter({ children, className, primaryButtonText, secondaryButtonText, ...rest }: CardFooterProps) {
+    return (
+        <div className={classNames(componentClasses.actionContainer, className)} {...rest}>
+            {primaryButtonText && (
+                <Button text={primaryButtonText} variant="primary" className={componentClasses.actionItem} />
+            )}
+            {secondaryButtonText && (
+                <Button text={secondaryButtonText} variant="secondary" className={componentClasses.actionItem} />
```
Vorbert-Kruk, line 29: There is no ability to pass any handlers to those buttons
As the button, without `click`  functionality, is just differently presented text, what do you think about adding some extra handlers? 🤔

src/components/dds/card/CardBody.tsx
```diff
@@ -0,0 +1,25 @@
...
+}
+
+/**
+ * Card component body
+ */
+export function CardBody({ children, className, description, ...rest }: CardBodyProps) {
+    return (
+        <div className={classNames(componentClasses.body, className)} {...rest}>
+            {description}
+            {children}
```
Vorbert-Kruk, line 20: Shouldn't we wrap `description` and `children` in some tags?
As `children` can be passed as a simple `string`, we can have a case when both strings are being visually merged together 

### Example
Code:
```typescript
                <Card>
                    <Card.Content>
                        <Card.Body description="Some description passed to 'description' prop.">
                            Children passed to Card.Body component
                        </Card.Body>
                    </Card.Content>
                </Card>
```

Results: 
image

### Conclusion

I don't know if we should be worried about this kind of behavior, that's why I just wanted to make sure that it is known 💪 
If we should strive towards adjusting it, we could put a simple `<span />` wrapper for the `description`, which would make the `children` to appear in a new line, instead of the same one 🎉

kubama (commented):
My only comment is that we should now be using `getRootClassNames` when setting class names for top level element. Here, it should be sufficient to use it in `Card` only, as other components are subcomponents and relevant styling will be inherited.

PatrycjaKubiczek (dismissed):

src/components/dds/card/CardFooter.tsx
qooban: Right, I missed that. Actually I was thinking of removing support for the buttons as with children you can add whatever you want. But for now I just changed props to allow passing all Button props for primary and secondary button.

src/components/dds/card/CardBody.tsx
qooban: I think it won't be very common to define both - `description` and `children`, but I added `span` wrapper.

qooban:
> My only comment is that we should now be using `getRootClassNames` when setting class names for top level element. Here, it should be sufficient to use it in `Card` only, as other components are subcomponents and relevant styling will be inherited.

@kubama Added it, but I got an issue with `Card.Footer`:

image

`dds__card__footer__item` class is shadowed by `dds__reboot-wrapper` class and the margins between buttons are set to 0px. Any ideas how to solve it? (it's not critical as for now I don't see button usecases for card footer in our mockups, but it would be good to have a way of overriding margins settings comming from reboot wrapper I think)

kubama:
Try moving `getRootClassNames` import to the beginning of the file

qooban:
Decided to use workaround (added wrapper for card footer items). I had to fix "With Additional Content" story as well, because `Typography` component doesn't play well with `Card` (specific `p` styling for cards overridden `Typography` styling).

kubama (approved):

src/components/dds/card/CardFooter.tsx
Vorbert-Kruk: Oh that idea seems to be a way to go, but probably in the future 🎉 
For now, taken approach seems to be good enough 💪

src/components/dds/card/CardBody.tsx
Vorbert-Kruk: Yeah, I could agree on that 😁 
Just wanted to raise the potential issue

But as it was covered, then we won't need to worry about it anymore 🎉

Vorbert-Kruk (approved):
LGTM 🎉

PR #228, opened 2023-05-17
RND-460 Create Notification component

PatrycjaKubiczek:
## Description

RND-460 Create Notification component

## Screenshots / Videos

Screenshot 2023-05-18 095328

## Checklist

## Tests

 

CI checks are green ✔️ 

## Documentation

n/a


src/components/dds/Notification/Notification.tsx
```diff
@@ -0,0 +1,134 @@
...
+                        <span className={componentClasses.headerTitle}>{title}</span>
+                        {closeIcon && (
+                            <StyledCloseButton
+                                variant="tertiary"
+                                className={classNames(componentClasses.button.base, componentClasses.closeButton)}
+                                onClick={() => setVisible(false)}
+                                size="sm"
+                                iconOnly
+                            >
+                                <Icon name="close-x" type="font-icon" className={componentClasses.icon} />
```
qooban, line 114: Optional:

Probably it will work as well:
```suggestion
                                <Icon name="close-x" className={componentClasses.icon} />
```

src/components/dds/Notification/Notification.tsx
```diff
@@ -0,0 +1,134 @@
+import * as React from 'react';
+import classNames from 'classnames';
+import styled from 'styled-components';
+import getRootClassNames from '../common/getRootClassNames';
+import { componentClasses } from 'dds-core/components/notification/notification';
+import { Button } from 'components';
```
qooban, line 6: This (at least so far) leads to problems when using components in `cloudify-marketplace`. If you could use relative paths for internal imports instead of `from 'components'` it would be great.

Ref. #231

src/components/dds/Notification/Notification.tsx
```diff
@@ -0,0 +1,134 @@
...
+    /**
+     * Primary action text for the notification
+     * */
+    primaryActionText: string;
+    /**
+     * Secondary action text for the notification
+     * */
+    secondaryActionText: string;
+    /**
+     * Update the number of Notifications visible on the Demo screen (Max of 3)
```
qooban, line 33: Demo screen? 🙂

src/components/dds/Notification/Notification.tsx
```diff
@@ -0,0 +1,134 @@
...
+     * */
+    timeStamp: string;
+    /**
+     * Primary action text for the notification
+     * */
+    primaryActionText: string;
+    /**
+     * Secondary action text for the notification
+     * */
+    secondaryActionText: string;
```
qooban, line 31: Maybe instead of having `primaryActionText` and `primaryAction` props, we could have `primaryActionButton` taking `ButtonProps`, so that user can actually define whatever needed?

I used that approach in `CardFooter`.

src/components/dds/Notification/Notification.tsx
```diff
@@ -0,0 +1,134 @@
...
+            data-dds="notification"
+        >
+            <div
+                className={classNames(componentClasses.base, visible && componentClasses.show)}
+                role="alert"
+                aria-live="polite"
+                aria-atomic="true"
+                aria-hidden={visible}
+            >
+                <Icon name={titleIcon} type="font-icon" className={componentClasses.icon} />
```
qooban, line 102: Optional:

`font-icon` is used by default, so you don't have to specify it.

src/components/dds/Notification/Notification.tsx
```diff
@@ -0,0 +1,134 @@
...
+                                onClick={() => setVisible(false)}
+                                size="sm"
+                                iconOnly
+                            >
+                                <Icon name="close-x" type="font-icon" className={componentClasses.icon} />
+                            </StyledCloseButton>
+                        )}
+                    </div>
+                    <p className={componentClasses.message}>{messageBody}</p>
+                    <div className="dds__notification__time-stamp">{timeStamp}</div>
```
qooban, line 119: if "dds__notification__time-stamp" is missing in `componentClasses`, maybe let's have:
```suggestion
                    <div className=`${componentClasses.base}__time-stamp`>{timeStamp}</div>
```
?

src/components/dds/Notification/Notification.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+        primaryActionText="View All"
+        primaryAction={() => console.log('Primary Action Clicked')}
+        secondaryActionText="Mark All as Read"
+        secondaryAction={() => console.log('Secondary Action Clicked')}
+        notificationCount={1}
+        titleIconType="icon"
+    />
+);
+export const noCloseIcon: NotificationStory = () => (
+    <>
```
qooban, line 31: Optional:

Wrapping tags can be removed.

src/components/dds/Notification/Notification.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+export interface NotificationProps extends React.HTMLAttributes<HTMLDivElement> {
+    /**
+     * Title text for the notification
+     * */
+    title: string;
+    /**
+     * Message body for the notification
+     * */
+    messageBody: string;
```
kubama, line 19:
```suggestion
    message: string;
```

src/components/dds/Notification/Notification.tsx
```diff
@@ -0,0 +1,134 @@
...
+    /**
+     * Time stamp for the notification
+     * */
+    timeStamp: string;
```
kubama, line 23: I think it is commonly used as a single world:
```suggestion
    /**
     * Timestamp for the notification
     * */
    timestamp: string;
```

src/components/dds/Notification/Notification.tsx
kubama: I like this idea :+1:

src/components/dds/Notification/Notification.tsx
```diff
@@ -0,0 +1,134 @@
...
+     * */
+    closeIcon: boolean;
+    /**
+     * Icon name for the notification
+     * */
+    titleIcon: string;
+    /**
+     * Icon type for the notification
+     * */
+    titleIconType: string;
```
kubama, line 47: I don't see this prop being used in the implementation.

src/components/dds/Notification/Notification.tsx
```diff
@@ -0,0 +1,134 @@
...
+     * */
+    notificationCount: number;
+    /**
+     * Whether the close icon should be visible
+     * */
+    closeIcon: boolean;
+    /**
+     * Icon name for the notification
+     * */
+    titleIcon: string;
```
kubama, line 43: There is a more specific type we could use here

src/components/dds/Notification/Notification.tsx
```diff
@@ -0,0 +1,134 @@
...
+        font-size: 1em;
+        line-height: 1;
+        && > i {
+            margin-right: 0;
+            font-size: 1em;
+    `;
+
+    return (
+        <div
+            className={classNames(getRootClassNames(className), componentClasses)}
```
kubama, line 90:
```suggestion
            className={getRootClassNames(className, componentClasses)}
```

src/components/dds/Notification/Notification.tsx
```diff
@@ -0,0 +1,134 @@
...
+            style={style}
+            {...rest}
+            data-dds="notification"
```
kubama, line 93: What do we need this for?

src/components/dds/Notification/Notification.tsx
```diff
@@ -0,0 +1,134 @@
...
+        line-height: 1;
+        && > i {
+            margin-right: 0;
+            font-size: 1em;
+    `;
+
+    return (
+        <div
+            className={classNames(getRootClassNames(className), componentClasses)}
+            style={style}
```
kubama, line 91: We can remove this line and L74 and let `style` to be spread as part of `...rest`

src/components/dds/Notification/Notification.tsx
```diff
@@ -0,0 +1,134 @@
...
+    style,
+    ...rest
+}: NotificationProps) {
+    const [visible, setVisible] = useState(true);
+
+    const StyledCloseButton = styled(Button)`
+        padding: 0;
+        font-size: 1em;
+        line-height: 1;
+        && > i {
```
kubama, line 83: We could reference `Icon` component in this selector instead of referencing its internals

src/components/dds/Notification/Notification.tsx
```diff
@@ -0,0 +1,134 @@
...
+    notificationCount,
+    closeIcon,
+    titleIcon,
+    titleIconType,
+    primaryAction,
+    secondaryAction,
+    style,
+    ...rest
+}: NotificationProps) {
+    const [visible, setVisible] = useState(true);
```
kubama, line 77: I can see `setVisible` is only called once with `false` arg. So once the notification is closed and the component gets hidden, how do you use it again? :)

src/components/dds/Notification/Notification.tsx
PatrycjaKubiczek: done: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/228/commits/67fc8954618e0dafc3fbde776ab900ae3af2c3df

src/components/dds/Notification/Notification.tsx
PatrycjaKubiczek: good to know, changed: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/228/commits/67fc8954618e0dafc3fbde776ab900ae3af2c3df

src/components/dds/Notification/Notification.tsx
PatrycjaKubiczek: copy-paste from Dell's docs 🙈 fixed: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/228/commits/67fc8954618e0dafc3fbde776ab900ae3af2c3df

src/components/dds/Notification/Notification.tsx
PatrycjaKubiczek: done: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/228/commits/67fc8954618e0dafc3fbde776ab900ae3af2c3df

src/components/dds/Notification/Notification.stories.tsx
PatrycjaKubiczek: removed: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/228/commits/67fc8954618e0dafc3fbde776ab900ae3af2c3df

src/components/dds/Notification/Notification.tsx
PatrycjaKubiczek: changed: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/228/commits/67fc8954618e0dafc3fbde776ab900ae3af2c3df

src/components/dds/Notification/Notification.tsx
PatrycjaKubiczek: removed: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/228/commits/67fc8954618e0dafc3fbde776ab900ae3af2c3df

src/components/dds/Notification/Notification.tsx
PatrycjaKubiczek: fixed: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/228/commits/67fc8954618e0dafc3fbde776ab900ae3af2c3df

src/components/dds/Notification/Notification.tsx
```diff
@@ -0,0 +1,124 @@
...
+        padding: 0;
+        font-size: 1em;
+        line-height: 1;
+        && > i {
+            margin-right: 0;
+            font-size: 1em;
+    `;
+
+    return (
+        <div className={getRootClassNames(className, componentClasses)} style={style} {...rest} data-dds="notification">
```
Vorbert-Kruk, line 84: I've didn't see `data-dds` attribute while reviewing other DDS component reviews
Do we need it? 🤔

src/components/dds/Notification/Notification.tsx
PatrycjaKubiczek: done: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/228/commits/67fc8954618e0dafc3fbde776ab900ae3af2c3df

src/components/dds/Notification/Notification.tsx
PatrycjaKubiczek: done: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/228/commits/67fc8954618e0dafc3fbde776ab900ae3af2c3df

src/components/dds/Notification/Notification.tsx
PatrycjaKubiczek: thanks for noticing it, changed: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/228/commits/1a6551e73051a000464be04db2f33c077d66eaa9

src/components/dds/Notification/Notification.tsx
PatrycjaKubiczek: removed: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/228/commits/1a6551e73051a000464be04db2f33c077d66eaa9

src/components/dds/Notification/Notification.tsx
PatrycjaKubiczek: changed: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/228/commits/1a6551e73051a000464be04db2f33c077d66eaa9

src/components/dds/Notification/Notification.tsx
PatrycjaKubiczek: I thought it could have some styling attached but not in the case of the notification component, so I removed it: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/228/commits/2f1e6080c3e9e0b3bffa973ce078087529abd0e5

src/components/dds/Notification/Notification.tsx
PatrycjaKubiczek: I thought it could have some styling attached but not in the case of the notification component, so I removed it: https://github.com/cloudify-cosmo/cloudify-ui-components/commit/2f1e6080c3e9e0b3bffa973ce078087529abd0e5

src/components/dds/Notification/Notification.tsx
PatrycjaKubiczek: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/228/commits/baad0720d7a220b1ed1d635c63638b6ecd5d143a changed

kubama (commented):
Thanks for addressing the comments. I added some more, but...
Actually at this point I also started thinking about use case for this component and... I'm not actually sure if we need it. The most obvious use case for it are toast notifications in the composer. But with current implementation these are made available not as components to render, but as a function to be called to display a toast which is... exactly what the vanilla implementation provides. And in this case a function interface as opposed to component interface makes much sense for a number of reasons. Logically such notifications are not part of the application component hierarchy, they are short-lived and can appear as a result of async operations. We may have multiple active notifications coming from different places in the code. In this case it's way easier to use a function call that displays such notification as opposed to mounting a component.
So, if we had it in a form of a component I think we would start using it by... providing a function that would display it on demand, possibly implementing a queue of self-destructing notifications. Do we need react component then?
@qooban @Vorbert-Kruk What do you think guys?

src/components/dds/Notification/Notification.tsx
```diff
@@ -16,21 +17,13 @@ export interface NotificationProps extends React.HTMLAttributes<HTMLDivElement>
...
      * */
-    primaryActionText: string;
+    timestamp: string;
     /**
-     * Secondary action text for the notification
-     * */
-    secondaryActionText: string;
-    /**
-     * Update the number of Notifications visible on the Demo screen (Max of 3)
+     * Update the number of Notifications visible on the screen (Max of 3)
```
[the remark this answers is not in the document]
kubama: Actually I reviewed the implementation and couldn't see any usages of this prop.

src/components/dds/Notification/Notification.tsx
```diff
@@ -40,90 +33,104 @@ export interface NotificationProps extends React.HTMLAttributes<HTMLDivElement>
...
+    padding: 0;
+    font-size: 1em;
+    line-height: 1;
+`;
+const StyledIcon = styled(Icon)`
+    && {
+        margin-right: 0;
+        font-size: 1em;
+    }
+`;
```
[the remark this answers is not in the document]
PatrycjaKubiczek: because otherwise we would have wrong styling: Screenshot 2023-05-22 084228

qooban:
> Do we need react component then?
> @qooban @Vorbert-Kruk What do you think guys?

I agree that for toast replacement we probably don't need react component, but...
Last week I asked Alex (he is the creator of RND-460) what for we need Notification component (in the context of DTW) and he said it's not to replace toast message, but to present error messages in login page 🙂

kubama:
> I agree that for toast replacement we probably don't need react component, but... Last week I asked Alex (he is the creator of RND-460) what for we need Notification component (in the context of DTW) and he said it's not to replace toast message, but to present error messages in login page slightly_smiling_face

Good to know. If so, I assume we are proceeding with this PR as planned.

Vorbert-Kruk:
> Do we need react component then? @qooban @Vorbert-Kruk What do you think guys?

Whole argument make sense to me.
If we'd have a toast-only use case for `Notfication` component, then it surely seems like a way to go.
But if we'd need to take other use cases into the account, then this approach could be not as practical.

Maybe it would be a good approach, to discuss it with Alex (as he's author of RND-460)
But as the work is in the late phase of development, It seems to me that it would be more safe to complete it, especially when we don't have any strict deadlines (AFAIK) for delivering this component

Vorbert-Kruk (approved):
Thanks for addressing comments
Looks ok to me 💪

src/components/dds/Notification/Notification.tsx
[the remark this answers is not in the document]
Vorbert-Kruk: For me it also looks suspicious as we should be able to implement DDS components as they come from dds components library

So maybe it would be worth to check if we are missing some styles from that place or if some global styles should be implemented in order to proceed with having correctly styled components, without any need of implementing additional styling

src/components/dds/Notification/Notification.tsx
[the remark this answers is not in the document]
PatrycjaKubiczek: removed: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/228/commits/393d6c4acf28cc5e111aa85b312b0a43729eb623

src/components/dds/Notification/Notification.tsx
```diff
@@ -40,90 +33,104 @@ export interface NotificationProps extends React.HTMLAttributes<HTMLDivElement>
...
+     * Secondary notification button props
+     *
+     * If not specified, the button will not be displayed
+     */
+    secondaryButton?: ButtonProps;
+    /**
+     * Whether the notification is visible
      * */
-    primaryAction: () => void;
+    visible?: boolean;
```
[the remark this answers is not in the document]
PatrycjaKubiczek: changed: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/228/commits/b82aac308b9a0d9c0cc717687cad54cc9230a83f

src/components/dds/Notification/Notification.tsx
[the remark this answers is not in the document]
kubama: Is this maybe caused by the `.dds__reboot-wrapper [something]` selector taking precedence over component specific stylesheets specified by just a class name?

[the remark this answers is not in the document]
PatrycjaKubiczek: .dds__reboot-wrapper doesn't affect the styling, it still displays wrong

src/components/dds/Notification/Notification.tsx
```diff
@@ -40,90 +33,104 @@ export interface NotificationProps extends React.HTMLAttributes<HTMLDivElement>
...
+    React.useEffect(() => {
+        setOpen(visible);
+    }, [visible]);
 
+    const handleOnClose = () => {
+        setOpen(false);
+        if (onClose) {
+            onClose();
+        }
+    };
```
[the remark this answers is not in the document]
PatrycjaKubiczek: do you think this affects the component in a negative way?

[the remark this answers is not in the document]
kubama: I do, IMHO its behavior is rather unexpected

src/components/dds/Notification/Notification.tsx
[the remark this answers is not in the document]
kubama: I checked the styling set for the button and I can see it's all being overridden:
image
So, having this styled component for the button has no effect.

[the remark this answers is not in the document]
kubama: Regarding style overrides for the icon, there are two issues making them required, both are relatively easy to fix.
1. There is a slight bug in component implementation - in L102 `className` should be set to `componentClasses.closeButtonIcon`, not `componentClasses.icon`. Once fixed it's no longer necessary to set `margin-right: 0`.
2. Correct font size as specified by `.dds__notification__close-button > .dds__notification__close-button__icon` gets overridden by classes coming from `Button` component, applied with selectors of equal specificity. This is due incorrect order of stylesheets inclusion. Looks like solution for this is to import all needed DDS stylesheets in a single file as opposed to importing them directly by their corresponding components source files. This way we will be able to control order of stylesheets inclusion independently of the order in which components are actually imported in the application.

Once these two issues are fixed, we should be able to get rid of `StyledIcon`.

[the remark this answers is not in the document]
kubama: And one more important thing - having styled components created in component's render function is a common anti-pattern leading to numerous issues - it creates a new component (not component instance) on each parent component render.

src/components/dds/Notification/Notification.tsx
```diff
@@ -0,0 +1,136 @@
...
+            {...rest}
+        >
+            <Icon name={titleIcon} className={componentClasses.icon} />
+            <div className={componentClasses.body}>
+                <div className={componentClasses.header}>
+                    <span className={componentClasses.headerTitle}>{title}</span>
+                    {closeIcon && (
+                        <StyledCloseButton
+                            variant="tertiary"
+                            className={classNames(componentClasses.button.base, componentClasses.closeButton)}
```
kubama, line 112: We don't need to set `componentClasses.button.base` here - it's already set by `Button`.
```suggestion
                            className={componentClasses.closeButton}
```

src/components/dds/Notification/Notification.tsx
```diff
@@ -0,0 +1,136 @@
...
+                </div>
+                <p className={componentClasses.message}>{message}</p>
+                <div className={`${componentClasses.base}__time-stamp`}>{timestamp}</div>
+                <div className={componentClasses.footer}>
+                    {secondaryButton && (
+                        <Button variant="secondary" className={componentClasses.button.base} {...secondaryButton} />
+                    )}
+                    {primaryButton && (
+                        <Button variant="primary" className={componentClasses.button.base} {...primaryButton} />
+                    )}
```
kubama, line 129:
```suggestion
                    {secondaryButton && (
                        <Button variant="secondary" {...secondaryButton} />
                    )}
                    {primaryButton && (
                        <Button variant="primary" {...primaryButton} />
                    )}
```

qooban (commented):
I see there are open discussions and I don't have more comments than already provided, so let me opt out.

src/components/dds/Notification/Notification.tsx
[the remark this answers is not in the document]
PatrycjaKubiczek: I replaced the scss styles with main.scss it helped a bit, but didn't fully resolve the issue - I was able to remove most of the custom styling - the only thing left is a font-size for close-button icon, IMO it's okay to have it, it's a minor thing.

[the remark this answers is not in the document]
PatrycjaKubiczek: updated: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/228/commits/003ad32d0d6470dcbf5319d9f4ec4810d266ffbf

src/components/dds/Notification/Notification.tsx
```diff
@@ -0,0 +1,112 @@
...
+export function Notification({
+    className,
+    title,
+    message,
+    timestamp,
+    closeIcon = true,
+    titleIcon,
+    primaryButton,
+    secondaryButton,
+    open,
```
qooban, line 66: I'm wondering - shouldn't we set the default here to `true`? 🤔 
As right now if I create new `Notification` component and add all required props, I won't see it. Am I correct?

src/components/dds/ActionMenu/ActionMenu.tsx
```diff
@@ -2,8 +2,7 @@ import React, { useEffect, useRef, useState } from 'react';
 import type { ButtonProps } from '../Button';
 import { Button } from '../Button';
 import { componentClasses } from 'dds-core/components/action-menu/action-menu';
-import 'dds-core/scss/base/_root.scss';
-import 'dds-core/components/action-menu/_action-menu.scss';
+import 'dds-core/scss/dds-main.scss';
```
qooban, line 5: I see changes from specific imports to 'dds-core/scss/dds-main.scss' (as a result of https://github.com/cloudify-cosmo/cloudify-ui-components/pull/228#discussion_r1202325481). 

I assume all DDS components look good, right? (as loki visual tests not complaining probably that's true)

qooban (approved):
Looks good. I haven't done very deep check.

src/components/dds/Notification/Notification.tsx
PatrycjaKubiczek: I was thinking that notification should appear only after some action, so that's why I haven't set `open` to `true` by default, but if you think I should change it to true, I'll change it.

src/components/dds/ActionMenu/ActionMenu.tsx
qooban: Good 👍🏻

src/components/dds/Notification/Notification.tsx
qooban: That's also fine, but then I think we should make `open` to be required, so that user will have to specify explicitly to show it or not. 

Right now the default behavior is that notification is not visible and there's no way to make it visible until you provide a prop which is optional. That may be misleading for component users.

PatrycjaKubiczek: sure thing, changed to required: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/228/commits/436075c6df1684174618ac5275fe5e7e34e016f6

qooban (approved):

PatrycjaKubiczek:
@kubama I forgot you'll be on sick leave since today, so I'll merge with 2 approvals, but if you'd have any comments about that I could cover them in separate PR

PR #233, opened 2023-05-24
RND-636 Initialized search component

Vorbert-Kruk:
## Checklist

- [ ] My code follows the [UI Code Style](https://cloudifysource.atlassian.net/l/c/x8fq902p).


src/components/dds/Search/Search.tsx
```diff
@@ -0,0 +1,116 @@
...
+import type { ButtonProps } from '../Button';
+import SearchActionButton from './SearchActionButton';
+import classNames from 'classnames';
+
+type ChangeEvent =
+    | React.ChangeEvent<HTMLInputElement>
+    | React.MouseEvent<HTMLButtonElement>
+    | React.KeyboardEvent<HTMLInputElement>;
+type HTMLInputProps = React.InputHTMLAttributes<HTMLInputElement>;
+export type SearchSize = keyof typeof componentClasses['sizes'];
```
Vorbert-Kruk, line 14: ### Note
I've tried to reference type names from other components, e.g. `Badge` 
While I was trying to export this file from `components/index.ts` it appears to be a problem we cannot export from the library multiple variables/types with the same name

The lesson is that in the future it may be advised to use more component-specific naming or to  encapsulate those types within namespaces

src/components/dds/Search/Search.tsx
```diff
@@ -0,0 +1,116 @@
...
+     * */
+    value?: string;
+
+    /**
+     * Function called on:
+     * - input value change
+     * - search icon click
+     * - enter key press
+     */
+    onChange: (value: string, event: ChangeEvent) => void;
```
Vorbert-Kruk, line 34: ### Note
In other component (e.g. `TextField`) this prop is optional
But as within our repos we are trying to utilize concept of controlled inputs, I've marked is as a required field

If you'd have a different suggestion, please make sure to start a discussion

src/components/dds/Search/Search.tsx
```diff
@@ -0,0 +1,115 @@
+import React, { useRef } from 'react';
+import { uniqueId } from 'lodash';
+import { componentClasses, defaultProps, base as componentPrefix } from 'dds-core/components/search/search';
+import 'dds-core/components/search/search.scss';
```
PatrycjaKubiczek, line 4: we should import the whole 'dds-core/scss/dds-main.scss' instead of partial scss because it affects the styling, see reference here: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/228#discussion_r1202325481

src/components/index.ts
```diff
@@ -9,6 +9,7 @@ export * from './dds/Masthead';
 export * from './dds/Notification';
 export * from './dds/TextInput';
 export * from './dds/Typography';
+export * from './dds/Search';
```
qooban: Minor thing, but for the sake of consistency - let's keep alphabetical order here:

```suggestion
export * from './dds/Notification';
export * from './dds/Search';
export * from './dds/TextInput';
export * from './dds/Typography';
```

src/components/dds/Search/Search.tsx
Vorbert-Kruk: Oh, I wasn't aware about such cases
Thanks for spreading that knowledge!
For sure I'll update the code 💪

Vorbert-Kruk: Code was adjusted within 8421e17709c6556e7ad962e58a2ddbe9f8652361 🎉

src/components/dds/Search/Search.tsx
```diff
@@ -0,0 +1,115 @@
...
+    className,
+    labelText,
+    style,
+    onChange,
+    value = '',
+    autoComplete = 'off',
+    size = 'lg',
+    ...inputProps
+}: SearchProps) => {
+    const displayClearButton = !!value;
```
qooban, line 55: `value` prop is optional, so if not specified, this variable will never be true, even though user provided search text.
Maybe worth to make `value` to be required the same as `onChange`?

src/components/dds/Search/Search.tsx
qooban: If `onChange` is required, I suspect `value` should also be required.

src/components/dds/Search/Search.tsx
qooban: I'll fix it right away in small PR.

Here it is: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/236. It bases on your PR.

README.md
```diff
@@ -82,9 +82,13 @@ There are few different types of tests in this package:
 1. Snapshot tests - `jest ./test/HtmlSnapshots.test.ts` (checks components HTML structure)
 1. Visual regression tests - `npm run test:loki` (takes screenshots of every story and compares them to reference images)
 
-NOTE: Visual regression tests depends on the environment - OS and browser used. Reliable results matching reference images can be gather only using Jenkins. In order to update screenshots: 
-* download and decompress `current.tar.gz` file taken from the Jenkins job artifacts
-* replace relevant screenshots in `.loki/reference` directory
+### Updating test artifacts
+1. Snapshot of components HTML  
+    Execute `npm run test:jest -- -u` command
```
qooban, line 87: There's more specific (and faster) way to just update snapshots:
`jest ./test/HtmlSnapshots.test.ts -u`

Maybe we should put it that way as in L82 we refer to that command.
Of course it requires `jest` to be available globally, but it can be easily fixed with prefixing the command with `npx `.

qooban (commented):
Looks good 👍🏻 Added some minor comments only.

src/components/dds/Search/Search.tsx
```diff
@@ -0,0 +1,115 @@
...
+                <input
+                    id={id}
+                    value={value}
+                    type="search"
+                    className={componentClasses.control}
+                    autoComplete={autoComplete}
+                    onChange={handleChange}
+                    onKeyDown={handleKeyDown}
+                    ref={inputRef}
+                    {...inputProps}
```
PatrycjaKubiczek, line 90: Optional:
add aria-label & placeholder when there's no labelText

[refrence](https://vanilla.delldesignsystem.com/2.28.0/index.html?path=/docs/components-search--basic):
_No label
When the search component doesn't contain a label, an aria-label and placeholder should be added to the search input._

README.md
Vorbert-Kruk: Oh, good to know!
Surely it seems like a nice idea to have a dedicated command for the plain purpose of updating snapshots
Having that said, you've gave me an idea 😁 

WDYT about having a dedicated `update-snapshots` command, which would be stored within `package.json`? 🤔 
Having that, we'd have a command (which we could reference in `README.md`) and in addition we would have additional place (`package.json`) where devs could look for clues related to updating snapshots

WDYT? 🤔

src/components/index.ts
Vorbert-Kruk: Sure 💪 

But in the future I wouldn't be worried about it, as it's not bringing any value from the perspective of components exposure
And having multiple components it may be time consuming to put every component in the right spot 😁

Vorbert-Kruk: Exports were adjusted within 7c7ef092bf7c46f8801318fbf46d086b57813109 🎉 
Funny part was that additional component was added, so I had to do a little investigation where to put `Search`, to match the alphabetical pattern 😄

src/components/dds/Search/Search.tsx
Vorbert-Kruk: If that wouldn't be problem, I'd leave it as it is
For now it's a dev choice what `placeholder` (and other properties like `aria-label`) should be put into `Search` 

Because of that, I'd interpret the referenced comment in the following way:
If during implementation of `Search` component would be decided to not provide `label` prop, `aria-label` and `placeholder` props should be provided instead

Something like a guideline for using this component 😉 

It may make sense also from the other perspective, which is the fact that from the component point of view it may be hard to decide what kind of value should be provided to `aria-label` and `placeholder` props, without knowing the use case of the component 😁 

Let me know if arguments above are making sense and if leaving the current interface would be okay from your side :pepe-salute:

README.md
qooban: Good idea 👍🏻 I would go this way.

src/components/dds/Search/Search.tsx
Vorbert-Kruk: According to our discussion, I've made the component to be uncontrolled - e518b97bd7b799d94d057578cb04498be1b7ff63

src/components/dds/Search/Search.tsx
PatrycjaKubiczek: sure thing, I left this comment as an optional :) so you'll make final decision

PatrycjaKubiczek (approved):

qooban (approved):

README.md
Vorbert-Kruk: Nice to hear 🎉 
Npm command was added within 61468270724edb42e7cc46b99c2a3100c4557c76

qooban (approved):

PR #241, opened 2023-05-30
RND-721 Add elevation utility function (DDS foundation)

qooban:
## Description

Added support for elevation classnames.

Ref.
https://www.delldesignsystem.com/foundations/elevation/
https://vanilla.delldesignsystem.com/2.28.0/index.html?path=/story/foundations-elevation--page

## Screenshots / Videos

Using different combinations of parameters provided to `getElevationClassName` produced:

image

## Checklist

## Tests
Manual tests only.

## Documentation
N/A


Vorbert-Kruk (approved):
LGTM 🎉

src/components/dds/common/elevation.ts
```diff
@@ -0,0 +1,16 @@
+import 'dds-core/scss/abstracts/helpers/_elevation.scss';
```
PatrycjaKubiczek, line 1: I wonder if we should add main.scss here too 🤔 I know that this is not really a component but I wonder if we could have some styling issues here due to css specificity or order 🤔

qooban: I checked if there are direct uses of `elevation` classes in `dds-core/src/scss/components`. There are no, so I guess in this case we should be good.

PatrycjaKubiczek (approved):

PR #244, opened 2023-06-01
RND-612 part 2

PatrycjaKubiczek:
## Description

RND-612 - update Login Page with MessageBar

## Screenshots / Videos

loginpageerror

## Checklist


src/components/dds/MessageBar/MessageBar.tsx
```diff
@@ -61,21 +62,24 @@ export function MessageBar({
...
             className={getRootClassNames(
+                className,
                 componentClasses.messageBar.base,
                 variant !== 'informative' && componentClasses.messageBar.colors[variant],
                 centerAlignment && componentClasses.messageBar.alignment.center,
                 globalLayout && componentClasses.messageBar.global,
                 !open && componentClasses.messageBar.hide
             )}
             aria-hidden={!open}
+            {...rest}
```
Vorbert-Kruk, line 82: ### Optional
As we are giving an user ability to override this `div` props, maybe we should assure that he's not overriding some fundamental ones like `role` and `aria-hidden`

For that, instead of doing:

```typescript
export interface MessageBarProps extends React.PropsWithChildren<React.HTMLAttributes<HTMLDivElement>> { 
```

we could do
```typescript
export interface MessageBarProps
    extends React.PropsWithChildren<Omit<React.HTMLAttributes<HTMLDivElement>, 'role' | 'aria-hidden'>> {
```

This would assure us that the user wouldn't be allowed to override required by the component prop values 💪 

WDYT? 🤔

Vorbert-Kruk (approved):
Looks good 💪

src/components/dell/DellLoginPage/DellLoginPage.tsx
```diff
@@ -168,6 +167,17 @@ export function DellLoginPage({
             </StyledWrapper>
             <StyledWrapper>
                 <form className={isLoginFormSubmitted ? componentClasses.form.validated : ''}>
+                    {error && (
+                        <MessageBar
+                            variant="error"
+                            className={getMarginClassName('bottom', 3)}
+                            dismissible={false}
+                            open={error !== null}
```
qooban, line 175: `error` can't be `null` in this line if I'm not mistaken, so maybe: 
```suggestion
                            open
```

src/components/dell/DellLoginPage/DellLoginPage.tsx
```diff
@@ -120,32 +119,32 @@ export function DellLoginPage({
...
             return;
         }
 
         const query = parse(location.search);
         const redirect = query.redirect as string;
 
-        onLogin(username, password, redirect).catch((loginError: Error) => {
-            console.error(loginError);
+        onLogin(username, password, redirect).catch(loginError => {
+            setError(loginError.message);
```
qooban, line 137: Are we sure we'll always get `message` field in the error? 🤔

src/components/dell/DellLoginPage/DellLoginPage.tsx
```diff
@@ -62,8 +63,6 @@ export interface DellLoginPageProps {
             | 'login'
             | 'error.missingCredentials'
             | 'error.incorrectCredentials'
```
qooban, line 65: I noticed that this key is never used. 
Shouldn't we have here `mapLoginError` function (which uses that key) similar to `LoginPage`? 🤔

src/components/dds/MessageBar/MessageBar.tsx
PatrycjaKubiczek: good idea, modified: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/244/commits/527707b5d2d95c48e2cbc74b4b33aa180508f4de

src/components/dell/DellLoginPage/DellLoginPage.tsx
PatrycjaKubiczek: changed it: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/244/commits/ad1a9b91ab8b1ef1b082745cee43c2c3f02c8dd5

src/components/dell/DellLoginPage/DellLoginPage.tsx
PatrycjaKubiczek: if the API haven't changed I think we should 🤔

Vorbert-Kruk (approved):
@qooban as you're not vacations and all of your discussions were addressed, I've agreed with @PatrycjaKubiczek to proceed further with this PR

[One discussion](https://github.com/cloudify-cosmo/cloudify-ui-components/pull/244#discussion_r1214243173) may have some room for development, but if that would be the case, I think that we should be good with doing so after your vacations 

Having that said, I'm approving this PR 🎉

PR #248, opened 2023-06-13
RND-829: Fixed inability to use components package in other repositories

Vorbert-Kruk:
## Description

Changes consists of removing removal of the `es` directory from the package root directory.
As it is being removed before the package is being published, I suspect that it may make a root cause of the issue
I've also got other ideas, but to confirm any of them I have to experiment a little bit, as I can't reproduce similar behavior which is present on `npm servers` after using `npm publish` command

## Checklist

## Tests

 

To test it I'll need to run `publish-version` script with a `preminor` version, to see if the package can be used now in other packages.

## Documentation

N/A


kubama:
> To test it I'll need to run publish-version script with a preminor version, to see if the package can be used now in other packages.

Can't you just build and test the package locally with the updated script? What difference does it make if preminor version is published?

Vorbert-Kruk:
> > To test it I'll need to run publish-version script with a preminor version, to see if the package can be used now in other packages.
> 
> Can't you just build and test the package locally with the updated script? What difference does it make if preminor version is published?

I've tested it locally, by trying to follow flow described within [Jenkinsfile](https://github.com/cloudify-cosmo/cloudify-ui-components/blob/master/jenkins/Jenkinsfile) and everything seems to be working as expected 💪 
But there is one difference, and the difference is that on the jenkins level we are just doing `npm publish`, without packing `cloudify-ui-components` beforehand
To test everything locally I've used `npm pack` command (instead of `npm publish`) and installed locally generated `.tar` file in other repos
As there may be some differences between `npm publish` and `npm pack` I'd be sure that the issue is resolved once the package would be successfully used from `npm` repository

For now I don't see much of a room for failure, but as there may be some differences, I'd like to test it out first, before releasing a new, stable version

kubama (approved):

kubama:
Makes sense :+1:

PR #249, opened 2023-06-15
RND-614 fix height issue on Dell Login Page

PatrycjaKubiczek:
## Description

- change height of Dell Login Page to fix issue with height

## Screenshots / Videos

Before:
there was an issue on some resolution: image
After:
Screenshot 2023-06-15 123743

## Checklist

## Tests

 

CI checks ✔️ 

## Documentation

n/a


src/components/dell/DellLoginPage/DellLoginPage.tsx
```diff
@@ -73,7 +73,8 @@ export interface DellLoginPageProps {
 
 const LoginContainer = styled.div`
     position: relative;
-    height: 50vh;
+    height: 50%;
+    min-height: 49vh;
```
kubama, line 77: Just thinking loud... but shouldn't the height be determined by this container content instead of being explicitly defined?

kubama: I removed both `height` and `min-height` and it seems to be all fine, or actually even better, because it doesn't cut out 'Edge Composer' header.

PatrycjaKubiczek: could you share a screenshot? this is what I got when I removed: height and min-height
Screenshot 2023-06-15 134201

kubama: Ah yes, it breaks indeed.

However I still think it shouldn't be necessary to control height of the top-level `div`. At the moment the layout is somehow hacky. Let alone the combination of `height: 50%` and `min-height: 49vh` is not intuitive and difficult to understand. At the same time child `div`s have both `height: 100%`. So the second child `div` is actually rendered outside of the containing top-level `div`. I think this layout should be fixed, preferably by using flex.

kubama:
```
const LoginContainer = styled.div`
    height: 100%;
    display: flex;
    flex-direction: column;
    color: ${colors.base.white()};
    input {
        width: 320px;
    }
    form {
        display: flex;
        flex-direction: column;
        align-items: center;
        justify-content: center;
    }
`;

const StyledWrapper = styled.div<{ $top?: boolean }>`
    flex: 1 0 330px;
    display: flex;
    flex-direction: column;
    align-items: center;
    justify-content: ${props => (props.$top ? 'flex-end' : 'center')};
    height: 100%;
    background-color: ${props => (props.$top ? colors.base.blue['blue-600']() : colors.base.white())};
`;
```

kubama: Edited: `height` can be now removed from `StyledWrapper` as it's now ignored.

PatrycjaKubiczek: okay, but I thought the design from mockups suggested the screen was supposed to be split in half - half blue, half white with a centered form and header 🤔  this solution is a bit different in design and layout 🤔

kubama: Isn't it how it works? The screen is split evenly across blue and white parts. The layout inside those parts remains unchanged.
image

PatrycjaKubiczek: it depends on what resolution you're looking :)
Screenshot 2023-06-16 094603

but since I heard we don't have to be pixel-perfect about mockups and design, I have something in between
I changed `flex: 1 0 500px` to have more space at the top: 
Screenshot 2023-06-16 094739
it looks good IMO, WDYT?

kubama: As for me, it looks fine :+1:.

PatrycjaKubiczek: great, implemented it: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/249/commits/c6040ba3bcd9e374648fb62da2a4945b09e23277 👍

kubama: However, this way you changed `flex-basis` for both parts actually, so the container is still split evenly. If you are testing in storybook then I suspect you may be not noticing the wrapping `div` with fixed height...

kubama: And what you are showing on the screenshot is the actual component plus some white space below it (which looks like if it was part of the component, but it actually isn't).

PatrycjaKubiczek: the specified height doesn't really affect it, see the screenshot from Stage: Screenshot 2023-06-16 100348

kubama: That's surprising... but if only it looks OK then fine :)

kubama (dismissed):

kubama (approved):

PR #250, opened 2023-06-16
RND-720 Drop dds-reboot-wrapper usage

kubama:
## Description
Dropped dds-reboot-wrapper usage for each individual DDS component and replaced it with globally applied dds-reboot stylesheet.
This changed the default font used, which affected many components - thus the number of updated loki images.

## Screenshots / Videos
N/A

## Checklist

## Tests
Updated loki images

 

## Documentation
N/A


Vorbert-Kruk (approved):
LGTM 💪

.storybook/theme.ts
```diff
@@ -1,4 +1,4 @@
-import { create } from '@storybook/theming';
+import {create} from '@storybook/theming';
```
qooban, line 1: Minor thing, but seeing that in other places we have white spaces in curly brackets, I'm wondering maybe ESLint configuration is not correct - not checking `.storybook` directory? 🤔

.loki/reference/chrome_laptop_Form_DatePicker_Basic.png
qooban, line 1: Looks like a good change 👍🏻 Previously the height of the time options was too big.

.loki/reference/chrome_laptop_Data_DataSegment_Action_Buttons.png
qooban, line 1: Maybe it won't be a problem, but the font size here is significantly bigger. Maybe worth to adjust it in overrides CSS? 🤔

.loki/reference/chrome_laptop_DDS_Select_Form_Validation.png
qooban, line 1: Looks like we didn't show asterisk before. Could you check if that's expected = bug fix?

.loki/reference/chrome_laptop_Cloudify_VisibilityField_Changeable.png
qooban, line 1: General:

Many icons (especially in `Visibility*` components) looks more cropped at the bottom. Will they look good in the applications (eg. Stage)? Maybe this is kind of loki issue? 🤔

src/components/index.ts
```diff
@@ -1,3 +1,7 @@
+import 'dds-core/scss/dds-reboot.scss';
+import 'cloudify-ui-common-frontend/styles/font-dds-Roboto.css';
+import './dds-overrides.css';
```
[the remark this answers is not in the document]
kubama: > I assume that L1 duplication with src/components/dds/index.ts is intended

Actually it was not intended at all... I think this duplication is not necessary, I'll remove it.

> I guess L3 is intended only for Stage and Composer, so it makes sense to keep it here.

Yes.

> I'm wondering about L2 - will we need to import Roboto font when using react-dds-components package manually to make it work? thinking Maybe it should be part of src/components/dds/index.ts as well?

You're right, I'll move it to `src/components/dds/index.ts`.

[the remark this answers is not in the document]
kubama: Done.

.storybook/theme.ts
kubama: Indeed, it was excluded by default and needed to be included explicitly, which I did along with applying needed fixes.

.loki/reference/chrome_laptop_DDS_Select_Form_Validation.png
kubama: We didn't show it before and we are not showing it now while it was present on the original screenshot. Looks like the difference threshold was not exceeded and the build was not failing despite of the difference. Nevertheless it is a bug which I'm going to fix in the upcoming PR as I will be extracting `Label` component to re-use it in `Dropdown`.

.loki/reference/chrome_laptop_Data_DataSegment_Action_Buttons.png
kubama: Applied

.loki/reference/chrome_laptop_Cloudify_VisibilityField_Changeable.png
kubama: Indeed, icon glyph is now rendered slightly outside of its container, which is why the loki image is cropped. However I checked in stage/composer and didn't notice it to make any actual difference, so I think we don't need to bother.

.loki/reference/chrome_laptop_DDS_Select_Form_Validation.png
qooban: OK, thanks!

.loki/reference/chrome_laptop_Cloudify_VisibilityField_Changeable.png
qooban: Great, thanks for doing additional checks.

qooban (approved):

PR #254, opened 2023-06-26
RND-464 Add DDS `Dropdown` component

kubama:
## Description
Added DDS `Dropdown` component

## Screenshots / Videos
image

## Checklist

## Documentation
Added JSDocs


src/components/dds/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,275 @@
...
+     */
+    required?: boolean;
+    /**
+     * Whether multiple options can be selected
+     */
+    multiple?: boolean;
+    /**
+     * Current dropdown values. Should be single element array when `multiple` is `false`
+     */
+    values?: Value[];
```
qooban, line 42: With `values` defined as array, user will have to provide that prop using an array when `multiple` is set to `false`, right?

I don't know would it help, but maybe you could consider discriminated union here, sth like:
```
{ multiple: false, value: Value, ... } | { multiple: true, value: Value[] }
```

src/components/dds/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,275 @@
...
+              /**
+               * Array of dropdown options
+               */
+              options: DDSDropdownOption<Value>[];
+          }
+        | {
+              /**
+               * Array of grouped dropdown options
+               */
+              optionsGroups: { title: string; options: DDSDropdownOption<Value>[] }[];
```
qooban, line 58: What if we have dropdown configured with groups and the same value for an option is used in at least two groups? 
Will we be able to distinguish between them? (e.g. in `onChange` function)

src/components/dds/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,275 @@
+/* eslint-disable react/jsx-no-bind */
+import classNames from 'classnames';
+import { componentClasses } from 'dds-core/components/dropdown/dropdown';
+import { componentClasses as tagClasses } from 'dds-core/components/tag/tag';
+import * as React from 'react';
+import { useEffect, useRef, useState } from 'react';
+import _, { flatMap, head, isNil, sum, uniqueId, xor } from 'lodash';
```
qooban, line 7: Currently `react-dds-components` does not have `lodash` in deps list, so you probably have to update `dds-package/package.json*` files. Probably having it in `peerDependencies` will be good enough.

src/components/dds/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,275 @@
...
+                        >
+                            <button
+                                type="button"
+                                title={currentValues.map(getMatchingOption).map(getOptionLabel).join(', ')}
+                            >
+                                {currentValues.length} selected
+                            </button>
+                            <button type="button" onClick={() => setCurrentValues([])}>
+                                <Icon name="close-x" className={tagClasses.closeButtonIcon} />
+                            </button>
```
qooban, line 209: Would it be possible to use `Button` component with some icons-related props instead of `button`?

src/components/dds/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,275 @@
...
+                        </div>
+                    )}
+                    <button className={componentClasses.chevron} type="button" onClick={toggleOptionsVisible}>
+                        <Icon name="chevron-down" />
+                    </button>
```
qooban, line 214: Same as above.

src/components/dds/PositionedContainer/usePositionedContainer.ts
```diff
@@ -0,0 +1,30 @@
...
+type GetPlacement = typeof getPlacement;
+type GetPlacementParams = Parameters<GetPlacement>[0];
+
+export default function usePositionedContainer<Trigger>(
+    visible: boolean,
+    placement: GetPlacementParams['placement'],
+    offsets?: GetPlacementParams['offsets']
+) {
+    const menuContainerRef = useRef<HTMLDivElement>(null);
+    const menuTriggerRef = useRef<Trigger>(null);
```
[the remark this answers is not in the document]
kubama: Not really - removed.

src/components/dds/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,275 @@
...
+import _, { flatMap, head, isNil, sum, uniqueId, xor } from 'lodash';
+import { Icon } from '../Icon';
+import { Label } from '../Label';
+import usePositionedContainer from '../PositionedContainer/usePositionedContainer';
+import { PositionedContainer } from '../PositionedContainer';
+import { placements } from 'dds-core/js/placement/placement';
+import { DropdownItem } from './DropdownItem';
+import { DropdownOptionItem } from './DropdownOptionItem';
+
+export interface DDSDropdownOption<Value> {
```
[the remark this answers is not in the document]
kubama: I think so. It should be working fine with virtually anything.

src/components/dds/Dropdown/Dropdown.tsx
kubama: No, I think it should be the component consumer responsibility to provide unique values. I consider the group title as a presentation-only attribute, as it is with `text`. In this case I think it should be working same as for non-groups case - we can have two options with same values but different texts, in which case it's not possible to distinguish between options once selected.

src/components/dds/Dropdown/Dropdown.tsx
kubama: Good idea, implemented.

Vorbert-Kruk:
I'll review it tomorrow 
Sorry for the delay

src/components/dds/Dropdown/Dropdown.tsx
kubama: Not really, it would result in some extra styling that applied that we don't want here.

src/components/dds/Dropdown/Dropdown.tsx
kubama: Same as above.

src/components/dds/Dropdown/Dropdown.tsx
qooban: OK

src/components/dds/Dropdown/Dropdown.tsx
qooban: OK

src/components/dds/Label/Label.tsx
```diff
@@ -0,0 +1,26 @@
...
+import React from 'react';
+import classNames from 'classnames';
+import { componentClasses as labelClasses } from 'dds-core/components/label/label';
+
+export interface LabelProps extends PropsWithChildren<React.LabelHTMLAttributes<HTMLLabelElement>> {
+    required?: boolean;
+}
+
+/**
+ * Label for form inputs, meant to be used internally
```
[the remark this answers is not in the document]
qooban: OK, sounds good.

Vorbert-Kruk (commented):
Component seems to be quite complex and there is no issue about it as it's not just a trivial as a plain button component 😄 
But looking at the amount of lines (almost 300) maybe we could do something to simplify it or to increase the readability of it

Wanted to share it as from my point of view if I'd need to change something in this component it would take some time for me to understand how this component is working
Maybe it's just a specific thing for this component, which has to be like that (as the component is quite complex)
Just wanted to share this observation as a feedback, which may bring some value/perspective 

As it's an overall comment, no action is required regarding it

src/components/dds/Dropdown/Dropdown.tsx
qooban: OK, makes sense. I missed that we have `value` and `text` for every option. In that case it's perfectly fine.

src/components/dds/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,279 @@
...
+    value: Value;
+    text?: string;
+    inactive?: boolean;
+}
+
+interface DDSDropdownIO<Value> {
+    /**
+     * Current dropdown value. Should be an array when `rest.multiple` is `true`
```
qooban, line 24: Minor thing - I'm not sure "rest." suits here.

src/components/dds/Dropdown/Dropdown.tsx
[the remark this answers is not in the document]
qooban: OK.

qooban (approved):

src/components/dds/Dropdown/Dropdown.tsx
kubama: Fixed

src/components/dds/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,279 @@
...
+        {
+            x: 0,
+            y: 4
+        }
+    );
+
+    useEffect(() => {
+        function outsideClickHandler(event: MouseEvent) {
+            const target = event.target as Node;
+            if (!menuTriggerRef.current?.contains(target) && !menuContainerRef.current?.contains(target)) {
```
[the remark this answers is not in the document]
kubama: Done

src/components/dds/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,279 @@
...
+
+    function getOptionsCount() {
+        return sum(normalizedOptions.map(group => group.options.length));
+    }
+
+    function getSelectAllButtonState() {
+        if (currentValues.length === 0) return false;
+        if (currentValues.length === getOptionsCount()) return true;
+        return 'mixed';
+    }
```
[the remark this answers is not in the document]
kubama: Done

src/components/dds/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,279 @@
...
+
+    function handleSelectAllClick() {
+        setCurrentValues(
+            currentValues.length === getOptionsCount()
+                ? []
+                : _(normalizedOptions).flatMap('options').map('value').value()
+        );
+    }
+
+    const tagVisible = (rest.multiple && !!currentValues.length) || undefined;
```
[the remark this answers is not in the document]
kubama: > In later phase this variable is being used as a boolean one

Actually it's being used as `true | undefined` - please see usage in L198 - having `boolean` as the type wouldn't allow this.

kubama:
@Vorbert-Kruk I extracted `DropdownMenu` component, hope it helps in terms of complexity.

src/components/dds/Dropdown/Dropdown.tsx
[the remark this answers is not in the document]
Vorbert-Kruk: Got it, you're right
In that case let's keep it as it is

PR #255, opened 2023-06-28
RND-466 Introduced DDS Pagination component

Vorbert-Kruk:
## Description

Created `Pagination` React component, basing on [vanilla implementation of DDS Pagination](https://vanilla.delldesignsystem.com/2.27.0/index.html?path=/docs/components-pagination--basic).

Please not that some functionalities were inspired by [vanilla implementation of pagination component](https://gitlab.dell.com/dao/dell-digital-design/design-language-system/systems/dds-2.0-components/-/blob/dev/src/components/pagination/pagination.js), as a measure helping in reflecting some of the component functionalities - I've left comments about it 💪 

### Basic
image

### With select opened
image

### With disabled buttons
image
image

## Checklist

## Tests

 

Added unit tests for the new component (6f42223d0f2123fb9d7f30ad3ee8241375691031).

## Documentation

Added JSDocs.


src/components/dds/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,214 @@
...
+    const [pageInputValue, setPageInputValue] = useState(currentPage.toString());
+
+    const handleItemsPerPageChange: SelectProps['onChange'] = value => {
+        const numericValue = parseInt(value as string, 10);
+        setPageSize(numericValue);
+    };
+
+    const getItemsRangeStart = () => {
+        return totalItems > 0 ? currentPage * pageSize + 1 - pageSize : 0;
+    };
```
Vorbert-Kruk, line 57: Idea behind this calculation was taken from [dds implementation](https://gitlab.dell.com/dao/dell-digital-design/design-language-system/systems/dds-2.0-components/-/blob/dev/src/components/pagination/pagination.js#L172)

src/components/dds/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,214 @@
...
+
+    const getItemsRangeEnd = () => {
+        const pageLastItem = currentPage * pageSize;
+
+        if (totalItems > 0) {
+            return pageLastItem > totalItems ? totalItems : pageLastItem;
+        }
+
+        return 0;
+    };
```
Vorbert-Kruk, line 67: Idea behind this calculation was taken from [dds implementation](https://gitlab.dell.com/dao/dell-digital-design/design-language-system/systems/dds-2.0-components/-/blob/dev/src/components/pagination/pagination.js#L177)

src/components/dds/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,214 @@
...
+        if (page < firstPage) {
+            return firstPage;
+        }
+
+        if (page > lastPage) {
+            return lastPage;
+        }
+
+        return page;
+    };
```
Vorbert-Kruk, line 79: Idea behind those conditions and calculations was taken from [dds implementation](https://gitlab.dell.com/dao/dell-digital-design/design-language-system/systems/dds-2.0-components/-/blob/dev/src/components/pagination/pagination.js#L204)

src/components/dds/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,214 @@
...
+    const handlePageInputBlur: PaginationRangeInputProps['onBlur'] = value => {
+        const numericValue = parseInt(value, 10);
+        const availablePage = getAvailablePage(numericValue);
+
+        if (matchers.number(numericValue) && availablePage !== currentPage) {
+            setCurrentPage(availablePage);
+        } else {
+            setPageInputValue(currentPage.toString());
+        }
+    };
```
Vorbert-Kruk, line 108: ### Note
Handling input changes and reflecting them on blur was inspired by 
[Vanilla implementation of DDS Pagination](https://vanilla.delldesignsystem.com/2.27.0/index.html?path=/docs/components-pagination--basic)
It can be seen that user should be able to enter input in any format and after blur all the validations and calculations are taken place 

More technical perspective can be seen [here](https://gitlab.dell.com/dao/dell-digital-design/design-language-system/systems/dds-2.0-components/-/blob/dev/src/components/pagination/pagination.js)

src/components/dds/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,214 @@
...
+
+    useEffect(() => {
+        onChange(currentPage, pageSize);
+    }, [currentPage, pageSize]);
+
+    return (
+        <div className={classNames(componentClasses.pagination.base, resizableClassName)} ref={paginationRef}>
+            <div className={componentClasses.summary}>
+                {/* eslint-disable-next-line jsx-a11y/label-has-associated-control */}
+                <label htmlFor={paginationPerPageId} className={componentClasses.perPageLabel}>
```
Vorbert-Kruk, line 130: ### Note
Label is implemented correctly, but eslint is giving us an error due to a fact that `select` is not wrapped within `label` element
As currently we have 2 similar occurrences across our codebase (2 additional are within this PR) maybe it would be worth to consider relaxing this rule

src/components/dds/Pagination/useResizableClassName.ts
```diff
@@ -0,0 +1,28 @@
...
+            setClassName(componentClasses.pagination.width768);
+        } else if (paginationWidth < width[1024]) {
+            setClassName(componentClasses.pagination.width1024);
+        }
+    };
+
+    useElementResize(paginationRef, handleClassNameChange);
+
+    return className;
+};
```
Vorbert-Kruk, line 26: ### Note
As mentioned on the [DDS Pagination documentation](https://www.delldesignsystem.com/components/pagination/?tab=Design#sizes), depending on the **component size**, different class name should be applied to it 

As there may be more components which would need a similar functionality, I've tried to be proactive and created `useElementResize` hook
This hook could be reused once similar functionality would be required in other components 🎉

src/components/dds/Pagination/useElementResize.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    useEffect(() => {
+        if (elementRef.current) {
+            resizeObserver.observe(elementRef.current);
+            return () => resizeObserver.unobserve(elementRef.current as HTMLElement);
+        }
+
+        return undefined;
+    }, [elementRef.current]);
+};
```
Vorbert-Kruk, line 25: ### Note
I've mirrored this functionality from [DDS Implementation](https://gitlab.dell.com/dao/dell-digital-design/design-language-system/systems/dds-2.0-components/-/blob/dev/src/components/pagination/pagination.js#L410), adding some react-like flavors

src/components/dds/Pagination/useResizableClassName.ts
```diff
@@ -0,0 +1,28 @@
...
+        } else if (paginationWidth < width[480]) {
+            setClassName(componentClasses.pagination.width480);
+        } else if (paginationWidth < width[640]) {
+            setClassName(componentClasses.pagination.width640);
+        } else if (paginationWidth < width[768]) {
+            setClassName(componentClasses.pagination.width768);
+        } else if (paginationWidth < width[1024]) {
+            setClassName(componentClasses.pagination.width1024);
+        }
+    };
```
Vorbert-Kruk, line 21: ### Note
This conditions was created basing on [DDS implementation](https://gitlab.dell.com/dao/dell-digital-design/design-language-system/systems/dds-2.0-components/-/blob/dev/src/components/pagination/pagination.js#L263)

src/components/dds/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,214 @@
...
+
+export interface DDSPaginationProps {
+    /**
+     * Total number of available items
+     */
+    totalItems: number;
+    /**
+     * Number of items displayed per page
+     */
+    itemsPerPage?: number;
```
qooban, line 24: If I'm not mistaken it's used only for initialization of `pageSize` state variable and that variable is responsible for keeping current value of page size select field. If that's correct, maybe more accurate would be to name it `initialPageSize`? 🤔

src/components/dds/Pagination/Pagination.tsx
qooban: According to https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/label-has-associated-control.md we can have `label` and input field as sibling, so we don't need wrapping, but this rule don't know that the input field is inside our `Select` component.

IMHO it's OK to disable the rule in this particular case.

Alternatively, you can try using `label` element that is defined inside `Select` component by providing `labelText` prop to `Select` component. For that you most probably will need some adjustments in `Select` component implementation.

src/components/dds/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,214 @@
...
+                    inactive={disablePreviousButton}
+                >
+                    <span className={componentClasses.prevPageLabel}>Previous</span>
+                </Button>
+                <div className={componentClasses.pageRange}>
+                    {/* eslint-disable-next-line jsx-a11y/label-has-associated-control */}
+                    <label htmlFor={pageRangeId} className={componentClasses.pageRangeLabel}>
+                        Page
+                    </label>
+                    <PaginationRangeInput
```
qooban, line 176: To avoid rule disabling, maybe you can try moving `<label>` element into `PaginationRangeInput` component?

That way `label`'s `htmlFor` prop and `input`'s `id` prop will be in the same file and maybe ESLint will be happier. 
Not only for satisfying ESLint rule it's worth trying, but also from logical perspective it's reasonable to have label input declared as part of the component providing that input.

src/components/dds/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,214 @@
...
+    /**
+     * Total number of available items
+     */
+    totalItems: number;
+    /**
+     * Number of items displayed per page
+     */
+    itemsPerPage?: number;
+    /**
+     * Function called when selected page or amount of items per page has changed
```
qooban, line 26: 'Amount of' is used for things you cannot measure, so I suggest to write "number of items per page" or just "page size" instead of "amount of items per page".

src/components/dds/Pagination/Pagination.utils.ts
```diff
@@ -0,0 +1,18 @@
+import type { SelectProps } from '../Select/Select';
+
+export const getPagesCount = (totalItems: number, itemsPerPage: number) => {
+    return Math.ceil(totalItems / itemsPerPage) || 1;
```
qooban, line 4: General:

In this PR sometimes I see `itemsPerPage`, and sometimes `pageSize`. Is it intentional to inform user about some differences between them or logically it is the same? If the latter, maybe let's stick to one name? 🤔

src/components/dds/Pagination/useElementResize.ts
```diff
@@ -0,0 +1,27 @@
+import { useEffect } from 'react';
+import type { MutableRefObject } from 'react';
+import { debounce } from 'lodash';
+
+type ResizeCallback = (elementOffsetWidth: number) => void;
+export type ResizableElementRef = MutableRefObject<HTMLElement | null>;
+
+const useElementResize = (elementRef: ResizableElementRef, onResizeCallback: ResizeCallback) => {
```
qooban, line 8: Can you check if maybe it would be possible to reuse/extract something from `useWidthObserver` hook? 🤔

src/components/dds/Pagination/useResizableClassName.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    useElementResize(paginationRef, handleClassNameChange);
```
qooban, line 23: I see from the code that what you need is element's width, so maybe instead of `useElementResize`, the hook we have - `useWidthObserver` would be suitable?

src/components/dds/Pagination/useResizableClassName.ts
```diff
@@ -0,0 +1,28 @@
+import { width, componentClasses } from 'dds-core/components/pagination/pagination';
+import { useState } from 'react';
+import useElementResize from './useElementResize';
+import type { ResizableElementRef } from './useElementResize';
+
+const useResizableClassName = (paginationRef: ResizableElementRef): string => {
```
qooban, line 6: It's probably a shortcut, but at first glance the name suggest that classname might be resizable ;-)  
Maybe `usePaginationClassName` as the implementation is tightly connected to `componentClasses.pagination`?

qooban (commented):
Sorry for late review, it took me a while.

src/components/dds/Pagination/useResizableClassName.ts
Vorbert-Kruk: Sure, seems like a nice suggestion
I'd just modify it slightly to be `usePaginationSizeClassName` to be more specific 🎉

src/components/dds/Pagination/useResizableClassName.ts
Vorbert-Kruk: It would be suitable
I did a little experiment and it seems that, looking at the implementation side of that hook, it would take some time to adjust component, so that it would be able to use it

As most likely we will migrate away from using this component in order to use its lit component counterpart, I'd leave it as it is to not spend more time on it 💪

src/components/dds/Pagination/useElementResize.ts
Vorbert-Kruk: I'd have a similar reasoning as mentioned [here](https://github.com/cloudify-cosmo/cloudify-ui-components/pull/255#discussion_r1250825387)

src/components/dds/Pagination/Pagination.tsx
Vorbert-Kruk: Oh, that' a nice tip, I didn't know it till now
Thanks for sharing it! 💪 

In that case I'll change it for sure 👌

src/components/dds/Pagination/Pagination.tsx
Vorbert-Kruk: I've given it a try and unfortunately this approach didn't satisfy eslint
image

From what I've read, eslint is not only being angry when `label` is not associated with a corresponding input, but also when the input is not a child of the `label`

Once `input` body was wrapped within `label` whole issue dissapears:
image

I've also did a quick experiment with removing `htmlFor` attribute from `label` wrapper, which resulted in getting an error:
image


Because of that, I'd leave it as it is
If in the future, we'd have more cases like that, we could relax this eslint rule, so that we wouldn't have a need of disabling it - maybe even there is a way of disabling just a wrapping requirement 🤔  

Let me know WDYT

src/components/dds/Pagination/Pagination.utils.ts
Vorbert-Kruk: Hmm, good point
Logically there are the same, the difference is that `itemsPerPage` can be consider a default `pageSize` multiplier
I'll unify them 💪

Vorbert-Kruk: Addressed within 170e535fe24b3d26f76234db5e8a4f06c904e629

src/components/dds/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,214 @@
+import React, { useEffect, useRef, useState } from 'react';
+import { componentClasses } from 'dds-core/components/pagination/pagination';
+import Select from '../Select/Select';
+import type { SelectProps } from '../Select/Select';
```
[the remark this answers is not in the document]
Vorbert-Kruk: In order to make this approach working, I had to add 
```typescript
export { default } from './Select';
```

line into the `Select/index.ts` file as I was getting the following error:
image

It shouldn't be concerning from the package built perspective, but I wanted you to know as it may make a beneficial piece of knowledge for the future 💪

src/components/dds/Pagination/Pagination.tsx
Vorbert-Kruk: I'm not sure about it, as it's also used for initializing select options with page sizes available within the component
Because of that I don't know if it would be better to leave it as `pageSize`

I've already changed it in 170e535fe24b3d26f76234db5e8a4f06c904e629, but looking at it right now I don't know if `itemsPerPage` name  was more accurate

As the purpose of this variable is to
- generate page size options
- initialize pagination with a default page size
It seems like an available choice

WDYT? Should it be modified or solution implemented within 170e535fe24b3d26f76234db5e8a4f06c904e629 seems to be okay? 🤔

src/components/dds/Pagination/Pagination.tsx
qooban: Strange, but OK, no time for digging deeper.

src/components/dds/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,214 @@
...
+const DDSPagination = ({ totalItems, pageSize: initialPageSize = 5, onChange }: DDSPaginationProps) => {
+    const paginationPerPageId = uniqueId();
+    const pageRangeId = uniqueId();
+    const paginationRef = useRef<HTMLDivElement | null>(null);
+    const resizableClassName = usePaginationSizeClassName(paginationRef);
+
+    const [currentPage, setCurrentPage] = useState(firstPage);
+    const [pageSize, setPageSize] = useState(initialPageSize);
+    const lastPage = getPagesCount(totalItems, pageSize);
+    const itemsPerPageOptions = getPageSizeOptions(totalItems, initialPageSize);
```
qooban, line 44: Maybe
```suggestion
    const pageSizeOptions = getPageSizeOptions(totalItems, initialPageSize);
```
to be fully inline?

src/components/dds/Pagination/Pagination.tsx
qooban: Hmm... Not sure. `pageSize` name for the prop and `pageSize` parameter name for `onChange` suggests that we have here fully controlled component and that we should use variable for `pageSize` initialized with value from`onChange` function. But IIUC it's not necessary, right? We can pass hardcoded number for that prop (currently named `pageSize`) and still use have that component working as expected (updating UI on changing page size select input and receiving updated `pageSize` value through `onChange` function).

Let's have a call if my description is not clear enough.

src/components/dds/Pagination/useResizableClassName.ts
qooban: OK, make sense, let's not spend more time on it.

src/components/dds/Pagination/Pagination.tsx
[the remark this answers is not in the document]
qooban: Wait so why is it possible to e.g. do `import { Icon } from '../Icon';` in `src/components/dds/Button/Button.tsx`? 🤔 

There's no need to `export { default } from './Button';`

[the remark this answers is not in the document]
qooban: Oh, because there are `{}` - brackets. Let's use this approach 🙂 I think it's more elegant than explicitly `export { default } from './Select';`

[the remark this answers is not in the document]
Vorbert-Kruk: Sure, in that case I'll go in that direction 💪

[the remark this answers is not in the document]
Vorbert-Kruk: Adjustment was done as a part of 991263c23b4402ba4d4bdb71b09d1a846484b434

src/components/dds/Pagination/Pagination.tsx
Vorbert-Kruk: Sure, seems like a nice idea 💪 
Applied as a part of 7aecc780e29b49131f58121760bfc641c4ca9ef5

src/components/dds/Pagination/Pagination.tsx
Vorbert-Kruk: After our discussion I've applied 2 changes:
- props adjustment (0b851c3759cfcff8d89b49311f9e1dc47ba9478c)
- initial execution of `onChange` handler (bb5679576ee179bf4b68d60e38400e5938f4c0e4)

qooban (approved):
Thanks for additional efforts 👍🏻

src/components/dds/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,220 @@
...
+     * Function called when selected page or number of items per page has changed
+     */
+    onChange: (page: number, pageSize: number) => void;
+}
+
+/**
+ * React implementation of DDS pagination component
+ * The component name is temporarily prefixed with 'DDS' due to name conflict with SemanticUI-based `Pagination` component.
+ */
+// TODO: Wybrać inicjalny pageSize
```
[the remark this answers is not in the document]
Vorbert-Kruk: Oh yeah, you're right 😄 
Removed within e913655642ed5d5e8f69d59e69a5edd71540235e

qooban (approved):

PR #273, opened 2023-08-09
RND-1002 update Readme.md

PatrycjaKubiczek:
## Checklist

- [ ] My code follows the [UI Code Style](https://cloudifysource.atlassian.net/l/c/x8fq902p).


src/components/ids/README.md
```diff
@@ -33,6 +33,30 @@ To upgrade `@edf/core` package:
 
 
 ### Development guidelines
+During development of IDS components these guidelines and best practices should be followed:
```
qooban, line 36: General:

1. I would align style in "Development guidelines" section. Before we had 3 main guidelines prefixed with `####`, with this addition there is additional ordered list with points. It might bring confusion to developers reading this. 
2. I would emphasize the changes to the guidelines we have right now. We were mainly talking about `args` that differentiates the current way from the new way and I think this is the most important aspect here. The other, especially more obvious don't need high visibility (they can e.g. be placed at the bottom).
3. Some groups are overlapping each other. For example points in "Props Clarity" and "Usage Guidance" fits also into "Clear Documentation" section, "Name and Purpose" are connected with "Naming conventions" we have. "Storybook Stories" are also somehow a way to document components, but the story definition is somehow worth to describe as it's connected to p.2. All in all I would consider regrouping and integrating with already existing structure.

I know these are general points, but I hope I provided valid reasoning. Let me know WDYT.
If more details needed, let's have a call.

PatrycjaKubiczek: @qooban I updated the README.md according to your suggestions :) let me know WDYT - should I change something more?

